PR #109, opened 2019-07-20
List extensions in docs, and add some helpful exports

charlex:
## Description

Export `ALL_GENERIC_EXTENSIONS`, `ALL_MARK_EXTENSIONS`, `ALL_NODE_EXTENSIONS`, and `ALL_EXTENSIONS` from `@remirror/core-extensions`. Tests the lengths of these arrays. Adds documentation for links.

## Checklist

- [x] I have read the [**CONTRIBUTING**](https://github.com/ifiokjr/remirror/blob/master/docs/contributing.md) document.

## Screenshot

Docs page:

image


ifiokjr:
@charlex I'm liking how you've structured this PR. It has tests, docs, and the code is nice and clean. I could learn a thing or two, especially, regarding writing docs for new functionality.

### Some thoughts

After thinking a bit more into the benefits of this new code I'm not sure it's enough to justify the addition to `remirror`.

There are currently two ways of creating a `RemirrorEditor`.

#### Method 1 - Use React

My preferred, but by no means required way, is to use a react tree for configuration.

```tsx
const EditorLayout: FunctionComponent = () => {
  return (
    <RemirrorManager>
      <RemirrorExtension Constructor={Bold} priority={1} />
      <RemirrorExtension Constructor={Italic} />
      <RemirrorExtension Constructor={Underline} />
      <ManagedRemirrorProvider
        attributes={{ 'data-testid': 'editor-instance' }}
        onChange={onChange}
        placeholder='Start typing for magic...'
        autoFocus={true}
        initialContent={EMPTY_PARAGRAPH_NODE}
      >
        <InnerEditor />
      </ManagedRemirrorProvider>
    </RemirrorManager>
  );
};
```

As you can imagine the lists of extensions doesn't really help much when configuring with this method.

#### Method 2 - Create your manager

Another way of configuring is via creating our own manager.

```tsx
const manager = ExtensionManager.create([
  ...baseExtensions,
  new CodeBlockExtension(),
  { extension: new PlaceholderExtension(), priority: 2 },
]);

const EditorLayout: FunctionComponent = () => {
  const onChange = () => console.log('I am magical 🧚🏾‍♀️');

  return (
    <RemirrorProvider
      manager={manager}
      attributes={{ 'data-testid': 'editor-instance' }}
      onChange={onChange}
      placeholder='Start typing for magic...'
      autoFocus={true}
      initialContent={EMPTY_PARAGRAPH_NODE}
    >
      <InnerEditor />
    </RemirrorProvider>
  );
};
```

In this case, yes, a list of extensions can be used quite easily but it does come with caveats. The order of extension is important and affects which functionality takes priority, this is why there is a priority prop when creating the manager. The lower the number the more important the extension 😋

It would be really difficult to create a list (as a part of the library) that captures order in a meaningful way for all possible users.

### Possible changes

A list could be useful for classifying extensions. However, right now each extension has a property with its type and we in the `ExtensionManager` this is frequently used to identify `Node`, `Mark` and `Plain` extensions.

I'd be open to adding a new property on `ExtensionManager.data` which keeps a record of the extension types. Something with the following shape

```ts
import { NodeExtension, Extension, MarkExtension } from '@remirror/core';

type ExtensionNameMap<GExtension extends AnyExtension> = {
  [name: string]: GExtension;
};

interface ExtensionRecord {
  node: ExtensionNameMap<NodeExtension>;
  mark: ExtensionNameMap<MarkExtension>;
  plain: ExtensionNameMap<Extension>;
}
```

I'd love to hear your thoughts 👍

charlex:
@ifiokjr Thanks so much for the thoughtful feedback! Tell you the truth, I didn't have a ton of value myself in consuming these new arrays, so I removed it from the PR to reflect my intent, and your feedback.

I'm not totally sure what you mean when you refer to putting a property on `ExtensionManager.data`.  Would this map just the extensions _loaded_? Or would it map _all_ extensions?

ifiokjr:
@charlex I'm happy to merge the new code once the lint issue is fixed.

If you take a look at the [failing pipeline](https://dev.azure.com/remirror/remirror/_build/results?buildId=330&view=logs) it's failing when checking the formatting of the edited file `docs/guides/terms/extension.mdx`. 

A quick fix would be to run `yarn fix` from the root of the repository. It will auto fix all prettier issues.

image

Once you've done that I'll merge 👍 

Regarding the proposed property on `ExtensionManager.data` I didn't really explain myself well. I was referring to the data that is initialized in the extension manager and wondering if it would be worthwhile adding a new property which held information on the extensions currently being used.

https://github.com/ifiokjr/remirror/blob/98cdbbd306195852ba0cf689d88668b9a4106e7c/%40remirror/core/src/extension-manager.ts#L148-L181

We can ignore this for now though 😄

PR #140, opened 2019-08-04
WIP: ESLint

charlex:
## Description

Replaces https://github.com/ifiokjr/remirror/pull/139

For https://github.com/ifiokjr/remirror/issues/134

Adds:

```
yarn lint:eslint
```

## Checklist

- [x I have read the [**contributing**][contributing] document.
- [x ] My code follows the code style of this project and `yarn fix` runs successfully.


.eslintrc.yml
```diff
@@ -0,0 +1,20 @@
+plugins:
```
ifiokjr: @charlex Is it possible to only use one eslint config for both the JavaScript and TypeScript files?

Currently, there is an eslintConfig in the root package.json as well and I'd rather one. 

Preference:

- Keep config in the package.json to prevent an overload of config files in the root directory.  
- If a root file is required then use a js file instead `.eslintrc.js`

benjie:
Thanks for your pull request, @charlex; we've integrated some of your suggestions into #150.

PR #151, opened 2019-09-20
test: add HTML snapshot tests

benjie:
I've added an HTML serializer for the Jest tests to make it easier to update the tests without having to go through and edit each one (in future). I've used prettier to format the HTML so it's easier to read, but I'm not yet sure that my indentation code is correct. I've also added `const FIRST_PARAGRAPH_SELECTOR = EDITOR_CLASS_SELECTOR + ' > p:first-child';` so that we get the content of the first paragraph since we don't care about the `<p>` tag itself in our tests; I'm not sure if this should be moved to a more central location like `@remirror/core` or not?

The test looks very similar to before, but it factors in more of the content and it can auto-update, so I'm calling that a win :grin: 

```diff
 it('should have a social editor', async () => {
      await $editor.focus();
      await $editor.type('This is text https://url.com');
-     await expect(innerHtml(EDITOR_CLASS_SELECTOR)).resolves.toInclude(
-       '<a href="https://url.com" role="presentation">https://url.com</a>',
+     expect($innerHTML(FIRST_PARAGRAPH_SELECTOR)).resolves.toMatchInlineSnapshot(
+       `This is text <a href="http://url.com" role="presentation">https://url.com</a>`,
      );
    });
```

If we need to use jest.retryTimes then we need to alias our inline snapshots, so hopefully we can avoid that.

### DIFF
--- a/e2e/helpers/index.ts
+++ b/e2e/helpers/index.ts
@@ -101,3 +101,15 @@ export const type = async ({ text, delay = 10 }: TypeParams) => page.keyboard.ty
 export * from './modifier-keys';
 
 export * from './images';
+
+interface HTMLObject {
+  _: 'HTML';
+  html: string;
+}
+
+function makeHtmlObject(htmlOrPromise: string | Promise<string>): Promise<HTMLObject> {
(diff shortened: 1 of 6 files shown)


ifiokjr:
Looks good 👍

PR #206, opened 2020-01-31
fix(core-extensions): make `keys()` return `KeyboardBindings`

ocavue:
## Description

First of all, excellent project. I really appreciate the idea of `Extension`, where I can put ALL my logic of one node/mark into a single class. Greet improvement for the readability.  When I using remirror in my project, I found a TypeScript issue as described below:

----

I want to extend those exist extensions in remirror but only change the key method, like something below:

```typescript
class MyCodeBlockExtension extends CodeBlockExtension {
  public keys(): KeyboardBindings {
    return {
      somekey: something
    }
  }
}
```

However, TS stop me from do this, beacuse `CodeBlockExtension.keys` return `{'Shift-Ctrl-\\': CommandFunction<any>}`. `MyCodeBlockExtension` must follow this type since it's a subcalss of `CodeBlockExtension`. 

I can prevent this error by 1) copy code from `@remirror` instead of import them; 2) use TS `Exclude` type to remove the definition of `CodeBlockExtension.keys`; 3) stop TS from type checking this class. 

But I think the best solution is to tell TS that what `keys` want to return is any bindings as well as it's a `KeyboardBindings` instance instead of some specific key bindings.  This is just an update for TS type hitting and it shouldn't affect any actual feature.

BTW, other `Extension` methods (like `commands`) should also have this issue. I can submit another PR to fix them if you agree to my opinion.

## Checklist

- [x] I have read the [**contributing**](https://github.com/ifiokjr/remirror/blob/master/docs/contributing.md)
      document.

## Screenshots

N/A

### DIFF
--- a/.changeset/spotty-bananas-remain.md
+++ b/.changeset/spotty-bananas-remain.md
@@ -1,0 +1,5 @@
+---
+'@remirror/core-extensions': patch
+---
+
+Make all built-in extensions' `keys()` return a `KeyboardBindings` object instead of a specific object (#206)
--- a/@remirror/core-extensions/src/extensions/history-extension.ts
+++ b/@remirror/core-extensions/src/extensions/history-extension.ts
@@ -1,4 +1,4 @@
-import { Extension, isFunction } from '@remirror/core';
+import { Extension, isFunction, KeyboardBindings } from '@remirror/core';
 import {
   BaseExtensionOptions,
   CommandFunction,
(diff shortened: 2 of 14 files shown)


.changeset/spotty-bananas-remain.md
```diff
@@ -0,0 +1,5 @@
+---
```
ifiokjr: Great work on adding this 👍

ifiokjr:
> BTW, other Extension methods (like commands) should also have this issue. I can submit another PR to fix them if you agree to my opinion.

I'll happily merge any future PR's.

ifiokjr:
@all-contributors please add @ocavue for code

Issue #215, opened 2020-02-06
Bad interaction between Italic Extension and mentions plugin


benjie:
## Description

Easiest way to reproduce:

1/ Edit fake-users.ts so one of the usernames contains an underscore, e.g. change `purplebird177` to `purple_bird177`:

https://github.com/remirror/remirror/blob/d7432539554b6a293fc0d92c29cc3d14189df0a9/%40remirror/showcase/src/data/fake-users.ts#L112

2/ `yarn build`

3/ Run the Next example: `cd examples/with-next && yarn dev`

4/ visit http://localhost:3000/editor/rich-social

5/ Auto-complete `@purple_bird177` (note the underscore) and then press `*`

**Expected**: text should read `@purple_bird177 *`

**Actual**: text reads `@purplbird177 `

This is with the rich editor but you can reproduce it with just the ItalicExtension:

```ts
  const extensions = useMemo(() => {
    return [{ extension: new ItalicExtension(), priority: 2 }];
  }, []);
```

**GIF**:

![Peek 2020-02-06 13-05](https://user-images.githubusercontent.com/129910/73940162-dca79380-48e2-11ea-9fc1-e373ebc8cca3.gif)

## Possible Fix

Don't have italic markdown match in the middle of words. I.e. the following should not match an underscore for italic purposes:

- `@_user`
- `@user_name`
- `user_name`
- `https://website.com/web_page`

I think it should also not match "special" text like links, mentions, tags; so the following should also not match for italic purposes:

- `@user_`
- `@__xuorig__`
- `#tag_`
- `http://site.com/_`

## Checklist

- [x] I have read the [**contributing**](https://github.com/remirror/remirror/blob/canary/docs/contributing.md) document.

ifiokjr:
Good spot

ifiokjr:
This still needs to be addressed. Again, I'll add a failing test case and then create a fix.

PR #217, opened 2020-02-06
feat(social): enable overriding mention and tag ids and hrefs

benjie:
## Description

Currently the code uses the username as the id; but users may edit their username over time so this could lead to broken content. A user's `id` never changes, so this PR allows you to store the ID you specify.

Similarly not every website users `/${username}` as the URL for a user; e.g. reddit uses `/u/${username}`. This PR allows overriding and storing custom id and href with mentions and tags.

## Checklist

### DIFF
--- a/@remirror/editor-social/src/social-types.ts
+++ b/@remirror/editor-social/src/social-types.ts
@@ -79,13 +79,16 @@ export type MatchName = 'at' | 'tag';
 export type MentionState = AtMentionState | HashMentionState;
 
 export interface UserData {
-  uid: string;
+  id?: string;
+  href?: string;
   username: string;
   displayName: string;
   avatarUrl: string;
 }
 
 export interface TagData {
+  id?: string;
+  href?: string;
   tag: string;
 }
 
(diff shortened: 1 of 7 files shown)


@remirror/editor-social/src/social-types.ts
```diff
@@ -79,13 +79,16 @@ export type MatchName = 'at' | 'tag';
 export type MentionState = AtMentionState | HashMentionState;
 
 export interface UserData {
-  uid: string;
+  id?: string;
```
benjie: I changed `uid` to `id`; this might be a breaking change (?). `id` for identifier seems more standard; `uid` is often interpretted as `user ID` rather than `unique ID`, so I thought since "identifier" already implies uniqueness `id` would be sufficient.

@remirror/showcase/src/data/fake-users.ts
```diff
@@ -109,7 +109,7 @@ export const fakeUsers: FakeUserObject = {
       email: 'jacinto.darosa@example.com',
       login: {
         uuid: '85d3419f-484b-48e2-b290-119d75d561de',
-        username: 'purplebird177',
+        username: 'purple_bird177',
```
benjie: Makes reproducing #215 easier; not necessary for this PR but seems harmless otherwise?

Issue #111, opened 2019-07-20
Feature: Populate <LinkInput /> with the current href


charlex:
## Description

Usecase: When a user clicks on an existing anchor link (`<a href="...">...</a>`), the `<LinkInput />` box should populate with the current value to allow for editing.

## Possible Implementation

Right now I am trying to implement this myself in my own custom link modal component, but I can't figure out how to get the current mark's attributes. Figuring this out right now, and may post an update later when I do. @ifiokjr Do you have any advice on how to do this?

charlex:
Progress update. I think I may have figured this out in my own custom implementation, but its kind of hacky. Posting here to see if this is the way it would be implemented within remirror itself:

```jsx

import { findDOMRefAtPos } from '@remirror/core';
import { useRemirror } from '@remirror/react';
import React, { useState } from 'react';
import { useToggle } from 'react-use';
import { Button, Modal } from 'reactstrap';

export const MyLinkButton = () => {
  // Context
  const { actions, manager, view } = useRemirror();
  const { selection } = manager.getEditorState();
  const removeLinkIsEnabled = () => actions.linkRemove.isEnabled();
  const updateLinkIsEnabled = () => actions.linkUpdate.isEnabled();

  // State
  const [href, setHref] = useState('');
  const [modalOpen, setModalOpen] = useToggle(false);

  return (
    <>
      <Button
        active={actions.linkUpdate.isActive()}
        disabled={!(updateLinkIsEnabled() || removeLinkIsEnabled())}
        onClick={() => {
          // WHEN USER CLICKS, 1) SEE IF THE NODE THEY CLICKED ON IS A LINK WITH AN ANCHOR
          // OR 2) OR SEE IF A CHILD IS AN ANCHOR,
          // AND THEN TAKE THE [HREF] AND SET IT TO THE VALUE TO ALLOW THE USER TO EDITI
          const result = findDOMRefAtPos(selection.from, view);

          if (result.nodeName === 'A' && result.hasAttribute('href')) {
            setHref(result.getAttribute('href').trim());
          } else {
            const foundChildAnchor = result.querySelector('a');
            if (
              foundChildAnchor &&
              foundChildAnchor.nodeName === 'A' &&
              foundChildAnchor.hasAttribute('href')
            ) {
              setHref(result.getAttribute('href').trim());
            }
          }

          // Open modal
          setModalOpen(true);
        }}
      />
      <Modal
        isOpen={modalOpen}
        value={href}
[...]

PR #223, opened 2020-02-11
feat(editor-wysiwyg): show current link value in LinkInput

tomas-c:
## Description

Solves #111 

## Checklist

### DIFF
--- a/.changeset/poor-lobsters-smoke.md
+++ b/.changeset/poor-lobsters-smoke.md
@@ -1,0 +1,5 @@
+---
+'@remirror/editor-wysiwyg': patch
+---
+
+Populate the Wysiwyg's `<LinkInput />` component with the current href to resolve the issue #111
--- a/@remirror/editor-wysiwyg/src/components/wysiwyg-menu.tsx
+++ b/@remirror/editor-wysiwyg/src/components/wysiwyg-menu.tsx
@@ -1,6 +1,6 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
-import { ActionNames, AnyFunction, Attrs, KeyOfThemeVariant } from '@remirror/core';
+import { ActionNames, AnyFunction, Attrs, getMarkAttrs, KeyOfThemeVariant } from '@remirror/core';
 import { bubblePositioner, useRemirrorContext } from '@remirror/react';
 import { useRemirrorTheme } from '@remirror/ui';
 import {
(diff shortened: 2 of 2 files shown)


tomas-c:
Checks fail due to "Check for vulnerabilities" task, not related to the changes in this PR

tomas-c:
Another alternative implementation could be to pass the current link attributes to the LinkExtension's activationHandler [here](https://github.com/remirror/remirror/blob/29c8ef4651ece848be7a70d944415c6c14c7c840/%40remirror/core-extensions/src/marks/link-extension.ts#L88)

ifiokjr:
> Checks fail due to "Check for vulnerabilities" task, not related to the changes in this PR

I've just updated the `yarn.lock` file so merging should be possible now. Everything looks good at a glance. I'll just take a look when I have a few more minutes before merging.

ifiokjr:
> Another alternative implementation could be to pass the current link attributes to the LinkExtension's activationHandler here

I like this idea in addition to what you've done.

ifiokjr:
Okay, I've rebased on top of the latest `canary`. If it passes I'll merge 😄

tomas-c:
> > Another alternative implementation could be to pass the current link attributes to the LinkExtension's activationHandler here
> 
> I like this idea in addition to what you've done.

@ifiokjr I've had a second look and realized that [LinkExtension](https://github.com/remirror/remirror/blob/29c8ef4651ece848be7a70d944415c6c14c7c840/%40remirror/core-extensions/src/marks/link-extension.ts#L88) is not the only one that calls the handler, it's also called by [MenuBar](https://github.com/remirror/remirror/blob/db20a731608970a4948fc97faf8c3a31bd6df333/%40remirror/editor-wysiwyg/src/components/wysiwyg-menu.tsx#L107) so putting active link attribute detection inside the first location is not enough. 

Options I could think of:

1. Leave this as it is now
2. Add current link attributes as PluginState which will require doing something like `const defaultLinkValue = manager.getPluginState<LinkExtensionState>('link')?.currentLinkAttrs?.href as string | undefined;` to get the current href
3. Add some sort of helper method / state selector that is exported from link-extension.ts that wraps `getMarkAttrs(state.newState, manager.schema.marks.link).href`

Option 2 unnecessarily adds a state variable and function call on every transaction. Option 3 in effect only changes the name of the function that has to be called from `getMarkAttrs` to something custom - not much benefit. 

I think it makes sense to leave this as it is unless there's a better way.

PR #232, opened 2020-02-15
fix(extension-code-block): backspace with content

### LINKED ISSUE #231: Backspace behaviour is wrong when a non-empty selection at the beginning of the code block
## Description

When I have something like below in my remirror editor (It's not the actual HTML but you can get my point):

```html
<h1>something before the code block</h1>
<code-block>
<start-selection>123<end-selection>456
</code-block>
```

When I press the Backspace key, I expect that the `123` part will be deleted. However, the actual behavior is that the cursor jumps to the block before the code block and nothing is been deleted because of [these codes](https://github.com/remirror/remirror/blob/v0.7.2/%40remirror/extension-code-block/src/code-block-extension.ts#L246-L247).

## Possible Fix

Just check if the selection is empty in `CodeBlockExtension.keys.Backspace`. I will submit a PR to fix it.

## Checklist

- [x] I have read the [**contributing**](https://github.com/remirror/remirror/blob/canary/docs/contributing.md) document.

ocavue:
## Description

Fixed #231 

## Checklist

## Screenshots

Before this PR:

before

After this PR:

after

### DIFF
--- a/.changeset/ninety-shirts-pump.md
+++ b/.changeset/ninety-shirts-pump.md
@@ -1,0 +1,5 @@
+---
+'jest-remirror': minor
+---
+
+`renderEditor` now accepts PrioritizedExtensions for more flexible testing.
--- a/.changeset/silly-rivers-smell.md
+++ b/.changeset/silly-rivers-smell.md
@@ -2,4 +2,4 @@
 '@remirror/extension-code-block': patch
 ---
 
-Doesn't steps into the previous node when pressing Backspace if the selection isn't empty
+Doesn't step into the previous node when pressing Backspace if the selection isn't empty
--- a/.changeset/silly-rivers-smell.md
+++ b/.changeset/silly-rivers-smell.md
@@ -2,4 +2,4 @@
 '@remirror/extension-code-block': patch
 ---
 
-Doesn't step into the previous node when pressing Backspace if the selection isn't empty
+Doesn't step into the previous node when pressing Backspace if the selection isn't empty #231.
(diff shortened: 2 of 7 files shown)


ifiokjr:
I really appreciate you adding the tests. I'll take a look at the change soon.

ocavue:
I write the test but I can't pass it because of some TypeScript error when running `yarn typecheck`. I try to understand what's happening but failed.

ocavue:
I found the problem! A `PrioritizedExtension` object can also be an `AnyExtension` object. So  TS can't to automatic type acquisition if a function requires `FlexibleExtensions` as a parameter.

In another word, a function parameter should only be  `PrioritizedExtension`  (like `ExtensionManager.create`) or only be `Extension` (like the current `renderEditor`).

So for now, I will not write a

ifiokjr:
I like the idea of updating `renderEditor` to accept the `PrioritizedExtension` argument. I think like you've mentioned it can go in a different PR possibly. I'll take a look around your codebase and see if I can figure out why they integration test is failing,

ifiokjr:
Not sure why this is failing the e2e as it's working on my machine.

ocavue:
That's so weird.

ocavue:
After pushing an empty commit, all tests have been passed.

ifiokjr:
I've added the `FlexibleExtension` functionality to jest-remirror and updated some wording. 

Really great work and understanding of the codebase. I'd love to see more from you.

PR #234, opened 2020-02-19
feat(core): improve keybinding composition

### LINKED ISSUE #204: Easy keybindings
## Description

It's quite common to want to take actions on certain keypresses (for example Enter -> save/send, Escape -> dismiss popover). The best way to do this in Remirror seems to be with an extension. However, the extensions seem to replace the keybindings rather than work together, it feels like Remirror extensions should automatically chain keybindings, or at least give the user a way of doing so. (I'm basing this on the implementation of BaseKeymapExtension which explicitly chains to baseKeymap; if I'm wrong maybe this code snippet can serve as an example anyway.)

## Example

Here's an example using the current API. I have not thought of a good name for `E` yet :wink: 

```ts
import { baseKeymap, chainCommands } from "prosemirror-commands";

export interface RemirrorEditorProps extends SproutEditorProps {}
type E = { commandModifier: boolean; shiftKey: boolean };
export interface SproutKeymapExtensionOptions extends BaseExtensionOptions {
  Enter?: (e: E) => boolean;
}

class SproutKeymapExtension extends Extension<SproutKeymapExtensionOptions> {
  name = "sproutKeymap";

  defaultOptions = {};

  // keys?(params: ExtensionManagerTypeParams<GType>): KeyboardBindings;
  public keys(): KeyboardBindings {
    const returnRule = (e: E) => (_state: any, _dispatch: any) => {
      return this.options.Enter
        ? this.options.Enter(e)
        : false;
    };
    return {
      Enter: chainCommands(
        returnRule({ commandModifier: false, shiftKey: false }),
        baseKeymap.Enter
      ),
      "Cmd-Enter": chainCommands(
        returnRule({ commandModifier: true, shiftKey: false }),
        baseKeymap.Enter
      ),
      "Shift-Cmd-Enter": chainCommands(
        returnRule({ commandModifier: true, shiftKey: true }),
        baseKeymap.Enter
      ),
      "Ctrl-Enter": chainCommands(
        returnRule({ commandModifier: true, shiftKey: false }),
        baseKeymap.Enter
      ),
      "Shift-Ctrl-Enter": chainCommands(
        returnRule({ commandModifier: true, shiftKey: true }),
        baseKeymap.Enter
      ),
      "Shift-Enter": chainCommands(
        returnRule({ commandModifier: false, shiftKey: true }),
        baseKeymap.Enter
      ),
    };
  }
}
```

Note this uses `chainCommands` to say that if the keypress is not handled it should pass onto the next handler; however we cannot determine what the next handler is so we just have to pass off to the default handler from `prosemirror-commands`. Maybe `keys()` should be passed the existing handlers?

## Checklist

- [x] I have read the [**contributing**](https://github.com/ifiokjr/remirror/blob/canary/docs/contributing.md)
      document.

ifiokjr:
## Description

Keys now use the new api for `CommandFunction` which provides a `next` method. This method allows for better control when deciding whether or not to defer to the `next` keybinding in the chain.

To override, create a new keybinding with another extension. Make sure the extension is created with a higher priority. The keybinding you create can either return true or false. By default if it returns true, no other keybindings will run. However if it returns `false` all other keybindings will be run until one returns `true`

`next` basically calls the every other keybinding regardless of what you return. It gives you full control of how the bindings are called.

**BREAKING CHANGE** - `CommandFunction` now takes in one parameter with the `dispatch`, `state`, and `view` properties. A new type `ProsemirrorCommandFunction` replaces the previous functionality. 

Closes #204

## Example

The following is an example where the enter key can be customised to ignore all lower priority `Enter` key bindings.

The `next` method allows full control beyond the return value. It allows both calling all lower priority key
bindings regardless of whether true or false has been called.

```tsx
import { BaseExtensionOptions, Extension, KeyBindings } from '@remirror/core';

interface CustomKeymapExtensionOptions extends BaseExtensionOptions {
  override: boolean;
}

export class CustomKeymapExtension extends Extension<CustomKeymapExtensionOptions> {
  /**
   * Shorthand method for creating the `BaseKeymapExtension`
   */
  public static of(options: CustomKeymapExtensionOptions) {
    return new CustomKeymapExtension(options);
  }

  get name() {
    return 'customKeymap' as const;
  }

  get defaultOptions() {
    return {
      override: false,
    };
  }

  /**
   * Injects the baseKeymap into the editor.
   */
  public keys(): KeyBindings {
    const { override } = this.options;
    return {
      Enter({ next }) {
        if (override) {
          // No other commands will be run
          return true;
        }
        next(); // Runs all lower priority commands
        return true;
      },
    };
  }
}
```

## Usage

To use the above example you could do the following.

```ts
import { ExtensionManager } from '@remirror/core';
import { baseExtensions } from '@remirror/core-extensions';
import { CustomKeymapExtension } from './custom-keymap-extension';

const manager = ExtensionManager.of([
  ...baseExtensions,
  { priority: 1, extension: CustomKeymapExtension.of({ override: true }) },
]);
```

This manager can now be used to create an editor where the `baseKeymap` for the `Enter` key is overridden.

If override is set to `false`, the method will call all lower priority keybindings while still allowing the behaviour you want for the extension.

## Checklist

### DIFF
--- a/.changeset/orange-swans-bathe.md
+++ b/.changeset/orange-swans-bathe.md
@@ -1,0 +1,15 @@
+---
+'@remirror/core-types': minor
+'@remirror/core': minor
+---
+
+- Change the signature of `CommandFunction` to only accept one parameter which contains `dispatch`, `view`,
+  `state`.
+- Create a new exported `ProsemirrorCommandFunction` type to describe the prosemirror commands which are still
+  used in the codebase.
+- Rename `KeyboardBindings` to `KeyBindings`. Allow `CommandFunctionParams` to provide extra parameters like
+  `next` in the newly named `KeyBindings`.
+- Create a new `KeyBindingCommandFunction` to describe the `Extension.keys()` return type. Update the name of
+  the `ExcludeOptions.keymaps` -> `ExcludeOptions.keys`.
+
+**BREAKING CHANGE**
--- a/.changeset/polite-tigers-sing.md
+++ b/.changeset/polite-tigers-sing.md
@@ -1,0 +1,6 @@
+---
+'jest-prosemirror': minor
+---
+
+Create a new class `ProsemirrorTestChain` for chaining the return from `createEditor`. Previously it was
+manually chained with a function. The plan is to extend this class within the `jest-remirror` codebase.
--- a/.changeset/pretty-apricots-hug.md
+++ b/.changeset/pretty-apricots-hug.md
@@ -1,0 +1,14 @@
+---
+'@remirror/core': minor
+---
+
+Improve the way `ExtensionManager` calls `Extension.keys` methods. Keys now use the new api for
+CommandFunctions which provides a `next` method. This method allows for better control when deciding whether
+or not to defer to the next keybinding in the chain.
+
+To override, create a new keybinding with another extension. Make sure the extension is created with a higher
+priority. The keybinding you create can either return true or false. By default if it returns true, no other
+keybindings will run. However if it returns `false` all other keybindings will be run until one returns `true`
+
+`next` basically calls the every lower priority keybinding in the extensions list. It gives you full control
+of how the bindings are called.
--- a/@remirror/core-extensions/package.json
+++ b/@remirror/core-extensions/package.json
@@ -33,6 +33,7 @@
     "@types/prosemirror-schema-list": "^1.0.1",
     "@types/prosemirror-state": "^1.2.3",
     "@types/prosemirror-view": "^1.11.2",
+    "map-obj": "^4.1.0",
     "prosemirror-commands": "^1.0.8",
     "prosemirror-dropcursor": "^1.2.0",
     "prosemirror-gapcursor": "^1.0.4",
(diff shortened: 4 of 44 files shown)


benjie:
Sweet!

Aside: the autogenerated docs make these kind of diffs humongous. Could we move the autogenerated docs to a different branch (e.g. `gh-pages`) and update them on merge to `master` via a GitHub action or similar? IMO the main repo should contain as little auto-generated content as possible.

ifiokjr:
> Aside: the autogenerated docs make these kind of diffs humongous.

100% agree. I've removed them for the time being.

> update them on merge to `master` via a GitHub action or similar

This is a great idea! I'll explore how to do so. If you have any suggestions / time to help out, I'm more than happy to merge any PRs.

benjie:
With the Graphile site we do autogeneration in travis:

https://github.com/graphile/graphile.github.io/blob/develop/.travis.yml

https://github.com/graphile/graphile.github.io/blob/develop/scripts/deploy

I'll probably move it to GitHub Actions at some point.

PR #236, opened 2020-02-20
feat(core): isEnabled via commands with dispatch=undefined

### LINKED ISSUE #226: isEnabled not implemented for bullet and ordered lists
## Description

If current selection is an ordered list, it's not possible to convert it to a bullet list by pressing the bullet list button. Yet the button is not disabled. This is because BulletListExtension and OrderedListExtension are using the default isEnabled implementation that always returns true.

## Possible Implementation

The following code solves this:

```
export class BulletListExtension extends NodeExtension {

  // <...> //

  public isEnabled({ getState, type, schema }: ExtensionManagerNodeTypeParams): CommandStatusCheck {
    return ({ command }) => {
      switch (command) {
        case 'toggleBulletList':
          return toggleList(type, schema.nodes.listItem)(
            getState(),
            undefined,
            (undefined as unknown) as EditorView,
          );
        default:
          return true;
      }
    };
  }

  // <...> //

}
```

Options:

1. Add the above isEnabled implementation to BulletListExtension and OrderedListExtension
2. Make the NodeExtension.isEnabled implementation always run the command without dispatch argument to determine if the command can be executed successfully (instead of always returning true).

I like the second option as:
* it's not possible to forget to implement isEnabled
* ensures that all other extensions correctly implement it

## Checklist

- [x] I have read the [**contributing**](https://github.com/remirror/remirror/blob/canary/docs/contributing.md) document.

tomas-c:
## Description

This PR changes code related to Extension.isActive() and Extension.isEnabled(). 

The action.isActive() and action.isEnabled() behaviour remains the same so this should not break code that doesn't access extensions directly.

### Change 1
Before this PR extensions by default would return true when calling Extension.isEnabled() and most extensions never implemented a custom Extension.isEnabled(). This resulted in issues such as #226 where action.isEnabled() would be true but command could not be run successfully. 

Extensions no longer need to implement isEnabled. It is provided by extension manager by calling commands with dispatch=undefined [as Prosemirror intended](https://prosemirror.net/docs/guide/#commands). It's now much more difficult to have action.isEnabled() return true unless a command can be successfully run. 

### Change 2

Extension.isActive previously allowed passing in a command name argument in the same way as Extension.isEnabled. I could not find any uses of this feature. I checked scrumpy/tiptap code and they only define a single isActive method per extension, not per action command. It seems to me that the main reason why isActive allowed passing in command names was because isEnabled allowed it too.

I got rid of this feature in this PR. It doesn't change the action.isActive() signature so if people later feel this was actually useful, it can be added in back later. Unless there was a reason for this that I missed?

### ~Change 3~

~History extension allowed to override getState() and dispatch methods which made it impossible to call a command with dispatch=undefined. I could not find any uses of this feature but if this is still needed I think a better way would be to have the extension manager provide this functionality for all extensions.~
 
## Checklist

Closes #226

### DIFF
--- a/@remirror/core-extensions/src/extensions/history-extension.ts
+++ b/@remirror/core-extensions/src/extensions/history-extension.ts
@@ -80,7 +80,7 @@ export class HistoryExtension extends Extension<HistoryExtensionOptions> {
       const { getState, getDispatch } = this.options;
 
       const wrappedState = isFunction(getState) ? getState() : state;
-      const wrappedDispatch = isFunction(getDispatch) ? getDispatch() : dispatch;
+      const wrappedDispatch = isFunction(getDispatch) && dispatch ? getDispatch() : dispatch;
 
       return method(wrappedState, wrappedDispatch, view);
     };
--- a/@remirror/core-extensions/src/marks/link-extension.ts
+++ b/@remirror/core-extensions/src/marks/link-extension.ts
@@ -2,7 +2,6 @@ import {
   Attrs,
   Cast,
   CommandMarkTypeParams,
-  CommandStatusCheck,
   ExtensionManagerMarkTypeParams,
   getMarkRange,
   getMatchString,
@@ -15,6 +14,7 @@ import {
   MarkExtensionSpec,
   MarkGroup,
   markPasteRule,
+  ProsemirrorCommandFunction,
   removeMark,
   selectionEmpty,
   updateMark,
@@ -99,31 +99,26 @@ export class LinkExtension extends MarkExtension<LinkExtensionOptions> {
       /**
        * A command to update the selected link
        */
-      updateLink: (attrs?: Attrs) => updateMark({ type, attrs }),
+      updateLink: (attrs?: Attrs): ProsemirrorCommandFunction => {
+        return (state, dispatch, view) => {
+          const { selection } = state;
+          if (selectionEmpty(selection) || (!isTextSelection(selection) && !isMarkActive({ state, type }))) {
+            return false;
+          }
+          return updateMark({ type, attrs })(state, dispatch, view);
+        };
+      },
       /**
        * Remove the link at the current position
        */
-      removeLink: () => {
-        return removeMark({ type, expand: true });
-      },
-    };
-  }
-
-  public isEnabled({ getState, type }: ExtensionManagerMarkTypeParams): CommandStatusCheck {
-    return ({ command }) => {
-      switch (command) {
-        case 'removeLink':
-          return isMarkActive({ state: getState(), type });
-        case 'updateLink': {
-          const { selection } = getState();
-          if (selectionEmpty(selection) || !isTextSelection(selection)) {
+      removeLink: (): ProsemirrorCommandFunction => {
+        return (state, dispatch, view) => {
+          if (!isMarkActive({ state, type })) {
             return false;
           }
-          return true;
-        }
-        default:
-          return true;
-      }
+          return removeMark({ type, expand: true })(state, dispatch, view);
+        };
+      },
     };
   }
 
--- a/@remirror/editor-wysiwyg/src/components/wysiwyg-menu.tsx
+++ b/@remirror/editor-wysiwyg/src/components/wysiwyg-menu.tsx
@@ -126,7 +126,7 @@ interface MenuItemProps extends Partial<WithPaddingProps> {
 const MenuItem: FC<MenuItemProps> = ({ state, onClick, Icon, variant, disabled = false, index }) => {
   return (
     <IconButton onClick={onClick} state={state} disabled={disabled} index={index}>
-      <Icon variant={variant} />
+      <Icon variant={variant} styles={{ color: disabled ? 'gray' : undefined }} />
     </IconButton>
   );
 };
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,7 @@
 # @remirror/docs
 
 ## 0.0.1
+
 ### Patch Changes
 
 - Updated dependencies [24f83413]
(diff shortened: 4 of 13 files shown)


@remirror/core-extensions/src/extensions/__tests__/history-extension.spec.ts
```diff
@@ -37,61 +36,3 @@ describe('commands', () => {
...
-      .shortcut('Mod-z')
-      .callback(content => {
-        state = content.state;
-      })
-      .callback(() => {
-        expect(mocks.getState).toHaveBeenCalledTimes(1);
-        expect(dispatcher).toHaveBeenCalledTimes(1);
-      });
-  });
-});
```
tomas-c: These tests were for HistoryExtension getState() and dispatch() override feature which could prevent calling commands with dispatch=undefined so I had to remove them. As I mentioned in the PR description, if this feature is needed, I suggest implementing it globally for all extensions via extension manager.

@remirror/core-extensions/src/extensions/history-extension.ts
```diff
@@ -108,27 +61,6 @@ export class HistoryExtension extends Extension<HistoryExtensionOptions> {
...
-        case 'undo':
-          return undoDepth(getState()) > 0;
-        case 'redo':
-          return redoDepth(getState()) > 0;
-        default:
-          return false;
-      }
-    };
-  }
-
```
tomas-c: This is not needed because redo() and undo() from prosemirror effectively do these checks when called with dispatch=false

@remirror/core-extensions/src/marks/link-extension.ts
```diff
@@ -99,31 +99,26 @@ export class LinkExtension extends MarkExtension<LinkExtensionOptions> {
       /**
        * A command to update the selected link
        */
-      updateLink: (attrs?: Attrs) => updateMark({ type, attrs }),
+      updateLink: (attrs?: Attrs): ProsemirrorCommandFunction => {
+        return (state, dispatch, view) => {
+          const { selection } = state;
+          if (selectionEmpty(selection) || (!isTextSelection(selection) && !isMarkActive({ state, type }))) {
```
tomas-c: You might notice that before these changes isEnabled used
```
selectionEmpty(selection) || !isTextSelection(selection)
```
but here I have
```
selectionEmpty(selection) || (!isTextSelection(selection) && !isMarkActive({ state, type }))
```

Turns out that the former would incorrectly result in isEnabled=false for this test case

@remirror/editor-wysiwyg/src/components/wysiwyg-menu.tsx
```diff
@@ -126,7 +126,7 @@ interface MenuItemProps extends Partial<WithPaddingProps> {
 const MenuItem: FC<MenuItemProps> = ({ state, onClick, Icon, variant, disabled = false, index }) => {
   return (
     <IconButton onClick={onClick} state={state} disabled={disabled} index={index}>
-      <Icon variant={variant} />
+      <Icon variant={variant} styles={{ color: disabled ? 'gray' : undefined }} />
```
tomas-c: With this you can see which extensions are disabled in the wysiwyg editor

tomas-c:
Rebased on latest canary, now all the build checks pass

tomas-c:
> I'm going to need to think about the idea of removing the getState and getDispatch from the `HistoryExtension`. It's not currently being used but was added to support sub editors.

I had a few thoughts here, tell me if they make sense:

1. Is it just the history extension that might need this feature or could others too? Would it make sense to come up with a global solution?
2. Would this be solved if extension manager provided access to the command, for example `action.command(state, dispatch, view)` in addition to `action()`, `action.isEnabled()` and `action.isActive()`?
3. Would this be solved if extension manager allowed to globally overwrite dispatch and getState method so that if you do overwrite it, all extensions still get the same state and dispatch functions? This would effectively move the wrapMethod logic from history extension to extension manager.

Is there an example implementation with a sub-editor? Would be happy to take a look and see if I can come up with other options. Difficult to reason about this without trying to implement a sub-editor. If there isn't an example, we could easily add this back in when it becomes needed. Perhaps then it will be clear what's the optimal way.

Also, I could also still make this PR work while keeping the wrapMethod like so:

```
private readonly wrapMethod = (method: ProsemirrorCommandFunction): ProsemirrorCommandFunction => {	
    return (state, dispatch, view) => {	
      const { getState, getDispatch } = this.options;	

      const wrappedState = isFunction(getState) ? getState() : state;	
      const wrappedDispatch = isFunction(getDispatch) && dispatch ? getDispatch() : dispatch;	

      return method(wrappedState, wrappedDispatch, view);	
    };	
  };
```

but then it feels wrong that this is the only extension which does this wraping

ifiokjr:
> Is there an example implementation with a sub-editor? Would be happy to take a look and see if I can come up with other options. Difficult to reason about this without trying to implement a sub-editor. If there isn't an example, we could easily add this back in when it becomes needed. Perhaps then it will be clear what's the optimal way.

There's no internal example, but here's [an example using prosemirror](https://prosemirror.net/examples/footnote/). 

> Would this be solved if extension manager allowed to globally overwrite dispatch and getState method so that if you do overwrite it, all extensions still get the same state and dispatch functions?

It's very specific to the history extension. Since you only ever want one history stack. Other extensions don't ever really need this.

I'm happy to merge this if you include that fix you mentioned to keep the history extension support for sub editors.

docs/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # @remirror/docs
 
 ## 0.0.1
+
```
tomas-c: Running `yarn fix` added this

tomas-c:
> It's very specific to the history extension. Since you only ever want one history stack. Other extensions don't ever really need this.
> 
> I'm happy to merge this if you include that fix you mentioned to keep the history extension support for sub editors.

Makes sense. I have rebased on canary and reverted those getState and dispatch changes.

PR #248, opened 2020-03-01
fix(core): Allow empty string default value for extraAttrs

ocavue:
## Description

`ExtraAttrsObject.default` should be a `sting`, `null` or `undefined` value. Based on the [definition](https://github.com/remirror/remirror/blob/c4645570d3e128d1decf3b9e904684263df7bf0c/%40remirror/core-types/src/base-types.ts#L224-L229), an attr is a required only if its `default` is  `undefined`. 

However, when `default` is an empty string (or `null`), this attr becomes a required unexpectedly. 

This PR fixs it.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
         env:
           CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
         with:
-          coverageCommand: 'echo Coverage'
+          coverageCommand: "echo Coverage"
           debug: false
 
       - name: build project
@@ -76,7 +76,7 @@ jobs:
 
       - name: notify slack
         uses: 8398a7/action-slack@v2
-        if: always() # Pick up events even if the job fails or is canceled.
+        if: github.ref == 'refs/heads/canary'
         with:
           status: ${{ job.status }}
           author_name: Remirror GitHub CI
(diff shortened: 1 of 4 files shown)


ifiokjr:
I like it! You'll need to rebase on top of `canary` I had pushed up some bad commits earlier today 😢

ocavue:
@ifiokjr After rebasing `canary`, GitLab Action still failed at job `notify slack`.

ocavue:
The error information is as below:

```
Run 8398a7/action-slack@v2
  with:
    status: Success
    author_name: Remirror GitHub CI
  env:
    CI: true
    GITHUB_TOKEN: ***
    SLACK_WEBHOOK_URL:  <---- this environment variable is empty!
```

Maybe we should only run this job for master branch.

ifiokjr:
> I think the reason is that SLACK_WEBHOOK_URL will only be available for collaborators. Maybe we should only run this job for the default branch, which can only be edited by collaborators.

Makes sense.

Ignore the failing chrome integration tests. For some reason there's one test that's become very flakey. I'll look at adding `jest-circus` to take advantage of the `jest.retryTimes(times: number)` method. 

Also, I guess I should look into why it's failing intermittently.

.github/workflows/ci.yml
```diff
@@ -76,7 +76,7 @@ jobs:
 
       - name: notify slack
         uses: 8398a7/action-slack@v2
-        if: always() # Pick up events even if the job fails or is canceled.
+        if: github.ref == 'refs/heads/canary'
```
ifiokjr: The problem with this is it will only pick up the event when it succeeds, which defeats the purpose.

I'll still merge, but I'll also update the config on the `canary` branch.

ifiokjr:
@ocavue I've [added](https://github.com/remirror/remirror/invitations) you as collaborator so you should be able to push directly.

ocavue:
Thank you.

PR #276, opened 2020-04-02
TypeScript should not perform ESLint's tasks

benjie:
## Description

When developing, it's common to comment out lines of code. When doing so, you shouldn't have to remove the import just for the development bundle to build. Similarly you may add arguments to a method that you're not using _yet_ but will be, or declaring variables again which you don't use right this second but will need. Detecting these issues are a job for ESLint; preventing a CI pass when these issues makes sense, but preventing you from even using your new code because it doesn't pass lint rules is a huge barrier to productivity.

This PR removes these lint-y rules from the base tsconfig, delegating them to ESLint.

## Checklist

- [ ] My code follows the code style of this project and `yarn fix` runs successfully.

### DIFF
--- a/support/tsconfig.base.json
+++ b/support/tsconfig.base.json
@@ -12,16 +12,17 @@
     "declaration": true,
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
-    "noUnusedLocals": true,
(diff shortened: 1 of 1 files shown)


ifiokjr:
Good point 👍

PR #281, opened 2020-04-13
feat(code-extension): support syntax highlight for upper language name

ocavue:
## Description

This simple PR add syntax highlight support for language name who has different case then the definition from Refractor. 

For example, Remirror support `yml` and `yaml` currently. With this PR, `YAML` and `YML` will also been supported.

As a proof of the necessity, GitHub website provides syntax highlight for all four language names below:

```yaml
language: "yaml"
```

```yml
language: "yml"
```

```YAML
language: "YAML"
```

```YML
language: "YML"
```

## Checklist

## Breaking changes

Removed  `isSupportedLanguage` and `getSupportedLanguagesMap`.

For `isSupportedLanguage`, I removed it just because it's unused in this repository. If you think we shouldn't break the backward compatibility easily, I can add this function back.

For `getSupportedLanguagesMap`, one reason is that it's never used in the repository, but the more important reason is the complex. Base on the new logic this PR provide, `getSupportedLanguagesMap` should return all possible word cases, which is a large amount of data with mostly garbage.

```
> getSupportedLanguagesMap([graphql])
{
	"GrApHqL": "graphql",
	"gRaPhQl": "graphql",
	"grAPhql": "graphql",
	...
}
```

### DIFF
--- a/.changeset/chatty-deers-guess.md
+++ b/.changeset/chatty-deers-guess.md
@@ -2,4 +2,4 @@
 "@remirror/extension-code-block": patch
 ---
 
-fix(code-extension): support syntax highlight for json5 and jinjia2.
+Support syntax highlight for `json5` and `jinjia2`.
--- a/.changeset/chatty-deers-guess.md
+++ b/.changeset/chatty-deers-guess.md
@@ -1,0 +1,5 @@
+---
+"@remirror/extension-code-block": patch
+---
+
+fix(code-extension): support syntax highlight for json5 and jinjia2.
--- a/.changeset/selfish-suits-look.md
+++ b/.changeset/selfish-suits-look.md
@@ -2,6 +2,6 @@
 "@remirror/extension-code-block": minor
 ---
 
-feat(code-extension): support syntax highlight for upper language names.
+Support syntax highlighting for uppercase language names.
 
 Removed `getSupportedLanguagesMap` and `isSupportedLanguage`.
--- a/.changeset/selfish-suits-look.md
+++ b/.changeset/selfish-suits-look.md
@@ -1,0 +1,7 @@
+---
+"@remirror/extension-code-block": minor
+---
+
+feat(code-extension): support syntax highlight for upper language names.
+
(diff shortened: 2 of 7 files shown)


@remirror/extension-code-block/src/code-block-extension.ts
```diff
@@ -181,7 +181,7 @@ export class CodeBlockExtension extends NodeExtension<CodeBlockExtensionOptions>
    * Create an input rule that listens converts the code fence into a code block with space.
    */
   public inputRules({ type }: ExtensionManagerNodeTypeParams) {
-    const regexp = /^```([a-zA-Z]*)? $/;
```
ocavue, line 184: `'``` '.match(/^```([a-zA-Z]*)? $/)` will return `{0: '``` ', 1: undefined, index: 0, input: '``` ', groups: undefined}`, which means the `language` variable is `undefined`. This could break something since the TypeScript assume that `language` is a `string`.

.changeset/chatty-deers-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remirror/extension-code-block": patch
+---
+
+fix(code-extension): support syntax highlight for json5 and jinjia2.
```
ifiokjr, line 5: This should be a human readable string.

```suggestion
Support syntax highlight for `json5` and `jinjia2`.
```

.changeset/selfish-suits-look.md
```diff
@@ -0,0 +1,7 @@
+---
+"@remirror/extension-code-block": minor
+---
+
+feat(code-extension): support syntax highlight for upper language names.
```
ifiokjr, line 5: Should be human readable.

```suggestion
Support syntax highlighting for uppercase language names.
```

packages/jest-prosemirror/package.json
```diff
@@ -50,7 +50,7 @@
     "test-keyboard": "^0.7.6"
   },
   "peerDependencies": {
-    "jest": "^24"
```
ifiokjr, line 53: Nice

PR #286, opened 2020-04-14
fix: do not inherit marks when insert hard break

### LINKED ISSUE #280: geMarkRange cannot handle hardBreak
## Description

I have created a link extension which trigger activation handler after click the link:

```
export class XLinkExtension extends LinkExtension {
  plugin({ type }: ExtensionManagerMarkTypeParams) {
    const that = this;
    return new Plugin({
      props: {
        handleClick(view, pos) {
          const { doc, tr } = view.state;
          const range = getMarkRange(doc.resolve(pos), type);

          if (!range) {
            return false;
          }

          // 选中链接的全部部分
          const $start = doc.resolve(range.from);
          const $end = doc.resolve(range.to);
          const transaction = tr.setSelection(new TextSelection($start, $end));

          view.dispatch(transaction);

          // 点击链接成功则激活链接处理
          that.options?.activationHandler?.();

          return true;
        }
      }
    });
  }
}

```

Most of the code is copied from the original link extension. 

After enabling the hard break, if the text with link is broken by a `<br/>`, then clicking the link will only select the first part of the link.

Kapture 2020-04-13 at 16 25 32

## Possible Fix

I have read the relative source code, but not familiar with the pm api, so currentlly cannot tell the key point.

## Checklist

- [ ] I have read the [**contributing**](https://github.com/remirror/remirror/blob/canary/docs/contributing.md) document.

zhaoyao91:
## Description

Fixes #280

## Checklist

- [ ] I have read the [**contributing**](https://github.com/remirror/remirror/blob/master/docs/contributing.md) document.
- [ ] My code follows the code style of this project and `yarn fix` runs successfully.

### DIFF
--- a/@remirror/core-extensions/src/nodes/hard-break-extension.ts
+++ b/@remirror/core-extensions/src/nodes/hard-break-extension.ts
@@ -28,7 +28,7 @@ export class HardBreakExtension extends NodeExtension {
   public keys({ type }: ExtensionManagerNodeTypeParams): KeyBindings {
     const command = chainCommands(convertCommand(exitCode), ({ state, dispatch }) => {
       if (dispatch) {
-        dispatch(state.tr.replaceSelectionWith(type.create()).scrollIntoView());
+        dispatch(state.tr.replaceSelectionWith(type.create(), false).scrollIntoView());
       }
 
       return true;


@remirror/core-extensions/src/nodes/hard-break-extension.ts
```diff
@@ -28,7 +28,7 @@ export class HardBreakExtension extends NodeExtension {
   public keys({ type }: ExtensionManagerNodeTypeParams): KeyBindings {
     const command = chainCommands(convertCommand(exitCode), ({ state, dispatch }) => {
       if (dispatch) {
-        dispatch(state.tr.replaceSelectionWith(type.create()).scrollIntoView());
+        dispatch(state.tr.replaceSelectionWith(type.create(), false).scrollIntoView());
```
ifiokjr, line 31: Interesting fix here 👍

ifiokjr:
Since PR's like this are very visual, a screenshot of what you've changed can be very helpful as well. 

It's not required, but it helps me to know that things are working at a glance.

zhaoyao91:
I will fix the change set and screenshot.

I encountered another case, and think maybe this issue still should be considred to be about the `getMarkRange`.

I have created a variable node extension before, which is a inline text placeholder. it could be within continuous marks like bold and link with surrounded text.

But #280 also happens to this situation:

Kapture 2020-04-15 at 10 39 45

So, is the target which should be fixed is the getMarkRange instead of hardBreak?

ifiokjr:
> I encountered another case, and think maybe this issue still should be considred to be about the `getMarkRange`.

Is it related to the hard break? It also seems like the variable node range could be at fault. 

I'm not sure I fully understand to be honest. Perhaps we should discuss in a seperate issue.

ifiokjr:
Closing as outdated.

PR #290, opened 2020-04-19
fix: image-extension `toDOM` can now safely access `this`

lanyusan:
## Description

image-extenison's `toDOM` is defined as object method, which couldn't pick up the extension's `this`, which would be a problem if `toDOM` needs to access `options` passed to the extension when it is setup.

This fix simply converts `toDOM` to arrow function to pick up `this` of extension.

## Checklist

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -157,6 +157,16 @@
       "contributions": [
         "bug"
       ]
+    },
+    {
+      "login": "lanyusan",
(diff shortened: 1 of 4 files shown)


lanyusan:
My pleasure. It is an easy fix. 

I was evaluating editors and came across `remirror`. I needed some pretty heavy customization so tried to customize the image extension thus found the problem.

PR #295, opened 2020-05-10
fix: remove `.step` property from collaboration steps

domq:
## Description

There doesn't seem to be any `.step` sub-structure in collaboration steps passed to e.g. [`onSendableReceived`](https://github.com/remirror/remirror/blob/000fdfb0d5e8da63e7ab8241ac4e6b8ccdc4ace0/%40remirror/extension-collaboration/src/collaboration-extension.ts#L120). Looking into the history of both Remirror and ProseMirror, I see that [Branch items can have a `.step`](https://github.com/ProseMirror/prosemirror-history/blob/master/src/history.js#L52), which might be where the confusion came from

## Checklist

(but on the other hand, neither was the old one.)

### DIFF
--- a/.changeset/eighty-trainers-change.md
+++ b/.changeset/eighty-trainers-change.md
@@ -1,0 +1,5 @@
+---
+'@remirror/extension-collaboration': patch
+---
+
+[fix] API mismatch in step enumeration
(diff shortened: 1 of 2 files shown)


ifiokjr:
@all-contributors please add @domq for code and bug reports.

PR #307, opened 2020-06-11
Playground overhaul

benjie:
## Description

- Import script to walk the exports of Remirror and add them to playground
- Updated to support latest Remirror API
- Add support for presets
- Small amount of refactoring
- Core extensions and presets are treated specially
  - Single concise list rather than one per module
  - Uses the simple name rather than adding the module path to the name
- Stub code for handling persistence across extension/preset changes (see GIF below)
- Bold/Italic buttons only render if the relevant extension is loaded
  - In future it would be great to auto-generate these command buttons for all extensions
- Tidied up the generated code

There's still lots to be done, but these things can be addressed in later PRs:

- Handle incorrect typescript definition fetches (pull the real type definitions in the import script, and load these in advance)
- Handle corruption on input after changing the supported extensions (pressing a key seems to trigger text deletion, bizarrely) - see `use-remirror-playground.tsx`
- Handle the issue when the cached document state is incompatible with the current schema (e.g. if it contains italic but you've removed the italic plugin)
- Handle extensions throwing errors (e.g. EpicModeExtension)
- Extend state persistence to the URL bar, so Remirror configurations and content states can be shared (e.g. to help people file an issue)
- Design and UX (e.g. adding a "copy" button, removing the code editor in basic mode, etc)

## Checklist

- [x] My code follows the code style of this project and `pnpm run fix` completed successfully.
- [ ] New code is unit tested and all current tests pass when running `pnpm run test` .

## Screenshots

Peek 2020-06-11 13-38


benjie:
Ready for review :+1:

PR #238, opened 2020-02-20
Realtime editing demo

benjie:
## Description

Adds an example of how to use realtime prosemirror editing using [yjs](https://github.com/yjs/yjs)'s [prosemirror support](https://github.com/yjs/y-prosemirror) and the [websocket transport](https://github.com/yjs/y-websocket).

I'd like some guidance on how best to achieve this to Remirror. It works, but it's pretty flaky.

## Checklist

- [x] I have read the [**contributing**](https://github.com/remirror/remirror/blob/master/docs/contributing.md) document.
- [x] My code follows the code style of this project and `yarn fix` runs successfully.
- [x] I have updated the documentation where necessary.
- [x] New code is unit tested and all current tests pass when running `yarn test` .

## Screenshots
Peek 2020-02-20 15-54

PR #319, opened 2020-06-25
feat: realtime collaborative editing

benjie:
## Description

This PR replaces #238; adding yjs support for the new Remirror. This can be demonstrated with the following code in the playground (advanced mode):

```ts
import React, { FC } from "react";
import { YjsExtension } from "remirror/extension/yjs";
import { RemirrorProvider, useManager, useRemirror } from "remirror/react";

const EXTENSIONS = [new YjsExtension()];

/**
 * This component contains the editor and any toolbars/chrome it requires.
 */
const SmallEditor: FC = () => {
  const { getRootProps } = useRemirror();

  return (
    <div>
      <div {...getRootProps()} />
    </div>
  );
};

const SmallEditorWrapper = () => {
  const extensionManager = useManager(EXTENSIONS);

  return (
    <RemirrorProvider
      manager={extensionManager}
    >
      <SmallEditor />
    </RemirrorProvider>
  );
};

export default SmallEditorWrapper;
```

As recommended by @dmonad in #238 we've switched over to `y-webrtc` so that the editing Just Works :tm: rather than requiring an external server.
 
## Checklist

- [ ] My code follows the code style of this project and `pnpm run fix` completed successfully.
- [ ] New code is unit tested and all current tests pass when running `pnpm run test` .

## Screenshots

Peek 2020-06-25 15-14

## Future tasks

- [ ] Make work as controlled component
- [ ] Allow configuring the provider (e.g. to use different channels)
- [ ] In playground, allow setting the channel as part of the URL/config so can copy URL and send to a collaborator.


benjie:
We have it working with other extensions too!!

Peek 2020-06-25 17-05

@remirror/playground/scripts/import-remirror.ts
```diff
@@ -93,6 +97,16 @@ export const INTERNAL_MODULES: Array<{ moduleName: string, exports: string[] }>
 `;
 }
 
+function forceTermination() {
+  const timeout = global.setTimeout(() => {
+    console.log(
+      "Look, I'm just a script, and far be it for me to tell you your job, dear human, but it seems to me that something has been keeping me alive for the last 5,000,000 nanoseconds (which feels like an eternity to me) since I completed my task. Maybe something opened a network connection? Who knows. Either way, it doesn't seem right, so I'm going to go ahead and exit.",
```
ifiokjr, line 103: 🤣

@remirror/extension-yjs/src/yjs-extension.ts
```diff
@@ -0,0 +1,156 @@
...
+  invariant,
+} from '@remirror/core';
+import { redo, undo, yCursorPlugin, ySyncPlugin, yUndoPlugin } from 'y-prosemirror';
+import { Doc } from 'yjs';
+import { WebrtcProvider } from 'y-webrtc';
+import { Observable } from 'lib0/observable';
+import { css } from 'linaria';
+
+/**
+ * yjs typings are very rough; so we define here the interface that we require
```
dmonad, line 17: This is fixed in y-webrtc@10.1.5 - I forgot to publish the tsc files.

It might be a good idea to remove the Observable import from lib0. Just because of this import, this extension won't run in node<12 because lib0 is an esm module. The modules in the Yjs repo are not affected because their bundle points to `lib0/dist/[module].cjs` which is a commonjs module.

dmonad: Well.. let me know if it is not fixed :wink: Trying to keep the typescript definitions up to date.

benjie: Awesome; Ifi was planning to contribute back some improved typings, but good to know they’re available soon! We need a “provider” type that all providers implement (webrtc, websockets), is there such a thing, @dmonad?

dmonad: That's a good idea. No there is not at the moment, but I'll add one in the next release https://github.com/yjs/yjs/issues/215

benjie: Awesome :raised_hands:

PR #404, opened 2020-07-29
feat(react-social): immediate emoji suggestions

### LINKED ISSUE #403: Consider allowing the EmojiExtension suggestion popup to appear after just the leading colon is typed
## Description

(May not be a bug but just a design choice.)

I'd like the `EmojiExtension` to show the popup as soon as I type a single colon `:`. Right now it doesn't appear until I type a letter after the colon.

The emoji list is not super useful with no letters typed unless you show recently-used first, but somehow I still prefer the feel of it showing me a popup as soon as I type the colon. Might just be me.

## Possible Fix

I looked into what was happening and initially thought I needed to set `matchOffset` for `prosemirror-suggest` to 0, but actually it seems more like the `PositioningExtension` responsible for the popup is returning the default offscreen position because it gets false when it calls `hasChanged()` when the popup first appears.

I don't really have any ideas for a fix yet.

I'm happy to help with any fix, but I don't yet understand the root issue well enough.

## Checklist

- [X] I have read the [**contributing**](https://github.com/remirror/remirror/blob/HEAD/docs/contributing.md) document.

ifiokjr:
## Description

Also add the character count to the `Indicator` component.

Closes #403 

## Checklist

### DIFF
--- a/packages/@remirror/react-social/src/components/social-editor-emoji.tsx
+++ b/packages/@remirror/react-social/src/components/social-editor-emoji.tsx
@@ -11,7 +11,7 @@ import {
   emojiSuggestionsItemStyles,
 } from '../social-styles';
 
-const EmojiDropdown = (props: Required<SocialEmojiState>) => {
+const EmojiDropdown = (props: SocialEmojiState) => {
   const { index, command, list } = props;
   const { focus } = useSocialRemirror();
   const { ref, top, left } = usePositioner('popupMenu');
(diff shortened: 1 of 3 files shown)


packages/@remirror/react-social/src/components/social-editor-emoji.tsx
```diff
@@ -67,10 +68,6 @@ const EmojiDropdown = (props: Required<SocialEmojiState>) => {
 export const SocialEmojiComponent = () => {
   const { index, list, command } = useSocialEmoji();
 
-  if (!command) {
```
ifiokjr, line 70: This was the issue with the delayed positioner. For the `positioner` to calculate the position it needs to have an element to inspect. So the element needs to be persisted.

PR #616, opened 2020-09-01
Add the persistent-selection extension

### LINKED ISSUE #601: Preserve text selection when editor loses focus (fake selection)
The editor should give the illusion that the current selection stays even when the editor lost focus. This should only happen for specific actions that keep the user in the mindset of the current selection. 

### Example use case

The user selects a piece of text and pressed the [link] menu button. When the "link" dialog opens, the editor will loose the focus and therefore also the text selection. This means that the user won't see any longer which text will be converted into a link.
image

### Solution idea

Atlassian seems to solves this be adding a fake-text-selection decoration at the place of the former selection:
image

---

This issue has been initially raised and discussed on [Discord](https://discord.com/channels/726035064831344711/726035065338986528/748847321416204403)

ankon:
### Description

Fixes #601 by adding a default extension to decorate the current selection, IFF the `persistentSelectionClass` builtin configuration is provided. 

### Checklist

### DIFF
--- a/.changeset/unlucky-pets-melt.md
+++ b/.changeset/unlucky-pets-melt.md
@@ -1,0 +1,9 @@
+---
+'@remirror/core': minor
+---
+
+Optionally allow to style the currently selected text
+
+This adds a new option for the builtin preset, `persistentSelectionClass`. If that is set to a valid CSS class name any selection in the editor will be decorated with this class.
+
+This can be used to keep an indication for the current selection even when the focus changes away from the editor.
--- a/packages/@remirror/core/src/builtins/__tests__/persistent-selection-extension.spec.ts
+++ b/packages/@remirror/core/src/builtins/__tests__/persistent-selection-extension.spec.ts
@@ -1,0 +1,53 @@
+import { extensionValidityTest, renderEditor } from 'jest-remirror';
+
+import { PersistentSelectionExtension } from '..';
+
+extensionValidityTest(PersistentSelectionExtension);
+
+describe('configuration', () => {
+  it.each([
+    {},
+    { persistentSelectionClass: undefined },
+    { persistentSelectionClass: '' },
+    { persistentSelectionClass: 'selection' },
+  ])('uses provided configuration (%j)', (configuration) => {
+    const {
+      add,
+      dom,
+      selectText,
+      nodes: { p, doc },
+    } = renderEditor([], { builtin: configuration });
+
+    add(doc(p('Some text')));
+    selectText({ from: 4, to: 9 });
+
+    expect(dom.outerHTML).toMatchSnapshot();
+  });
+});
+
+describe('focus', () => {
+  it('retains selection and decoration', () => {
+    const builtin = { persistentSelectionClass: 'selection' };
+    const editor = renderEditor([], { builtin });
+    const {
+      add,
+      dom,
+      selectText,
+      nodes: { p, doc },
+    } = editor;
+
+    add(doc(p('Some text')));
+    const range = { from: 4, to: 9 };
+    selectText(range);
+
+    // Verify that the selection has been applied
+    expect(editor.state.selection.from).toBe(range.from);
+    expect(editor.state.selection.to).toBe(range.to);
+
+    (editor.view.dom as HTMLElement).blur();
+
+    expect(dom.outerHTML).toMatchSnapshot();
+    expect(editor.state.selection.from).toBe(range.from);
+    expect(editor.state.selection.to).toBe(range.to);
+  });
+});
--- a/packages/@remirror/core/src/builtins/persistent-selection-extension.ts
+++ b/packages/@remirror/core/src/builtins/persistent-selection-extension.ts
@@ -1,0 +1,58 @@
+import { ExtensionPriority } from '@remirror/core-constants';
+import type { Static } from '@remirror/core-types';
+import { isNodeSelection } from '@remirror/core-utils';
+import { Decoration, DecorationSet } from '@remirror/pm/view';
+
+import { extensionDecorator } from '../decorators';
+import { PlainExtension } from '../extension';
+import type { CreatePluginReturn } from '../types';
+
+export interface PersistentSelectionOptions {
+  persistentSelectionClass?: Static<string>;
+}
+
+export const DEFAULT_PERSISTENT_SELECTION_CLASS = 'selection';
+
+@extensionDecorator<PersistentSelectionOptions>({
+  defaultOptions: {
+    persistentSelectionClass: DEFAULT_PERSISTENT_SELECTION_CLASS,
+  },
+  defaultPriority: ExtensionPriority.Medium,
+  staticKeys: ['persistentSelectionClass'],
+})
+export class PersistentSelectionExtension extends PlainExtension<PersistentSelectionOptions> {
+  get name() {
+    return 'persistentSelection' as const;
+  }
+
+  createPlugin(): CreatePluginReturn {
+    return {
+      props: {
+        decorations: ({ doc, selection }) => {
+          // Do not run the extension at all when there is no actual decoration
+          // to be done.
+          if (!this.options.persistentSelectionClass) {
+            return;
+          }
+
+          if (selection.empty) {
+            return;
+          }
+
+          const decorationAttrs = {
+            class: this.options.persistentSelectionClass,
+          };
+          let decoration: Decoration;
+
+          if (isNodeSelection(selection)) {
+            decoration = Decoration.node(selection.from, selection.to, decorationAttrs);
+          } else {
+            decoration = Decoration.inline(selection.from, selection.to, decorationAttrs);
+          }
+
+          return DecorationSet.create(doc, [decoration]);
+        },
+      },
+    };
+  }
+}
--- a/packages/@remirror/core/src/builtins/__tests__/__snapshots__/persistent-selection-extension.spec.ts.snap
+++ b/packages/@remirror/core/src/builtins/__tests__/__snapshots__/persistent-selection-extension.spec.ts.snap
@@ -1,0 +1,78 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`configuration uses provided configuration ({"persistentSelectionClass":""}) 1`] = `
+<div contenteditable="true"
+     role="textbox"
+     aria-multiline="true"
+     aria-label
+     class="ProseMirror remirror-editor"
+>
+  <p>
+    Some text
+  </p>
+</div>
+`;
+
+exports[`configuration uses provided configuration ({"persistentSelectionClass":"selection"}) 1`] = `
+<div contenteditable="true"
+     role="textbox"
+     aria-multiline="true"
+     aria-label
+     class="ProseMirror remirror-editor"
+>
+  <p>
+    Som
+    <span class="selection">
+      e tex
+    </span>
+    t
+  </p>
+</div>
+`;
+
+exports[`configuration uses provided configuration ({}) 1`] = `
+<div contenteditable="true"
+     role="textbox"
+     aria-multiline="true"
+     aria-label
+     class="ProseMirror remirror-editor"
+>
+  <p>
+    Som
+    <span class="selection">
+      e tex
+    </span>
+    t
+  </p>
+</div>
+`;
+
+exports[`configuration uses provided configuration ({}) 2`] = `
+<div contenteditable="true"
+     role="textbox"
+     aria-multiline="true"
+     aria-label
+     class="ProseMirror remirror-editor"
+>
+  <p>
+    Some text
+  </p>
+</div>
+`;
+
+exports[`focus retains selection and decoration 1`] = `
+<div contenteditable="true"
+     role="textbox"
+     aria-multiline="true"
+     aria-label
+     class="ProseMirror remirror-editor"
+>
+  <p>
+    Som
+    <span class="selection">
+      e tex
+    </span>
+    t
+  </p>
+</div>
+`;
--- a/packages/@remirror/core/src/builtins/builtin-preset.ts
+++ b/packages/@remirror/core/src/builtins/builtin-preset.ts
@@ -11,12 +11,16 @@ import { InputRulesExtension } from './input-rules-extension';
 import { KeymapExtension, KeymapOptions } from './keymap-extension';
 import { NodeViewExtension } from './node-views-extension';
 import { PasteRulesExtension } from './paste-rules-extension';
+import {
+  PersistentSelectionExtension,
+  PersistentSelectionOptions,
+} from './persistent-selection-extension';
 import { PluginsExtension } from './plugins-extension';
 import { SchemaExtension } from './schema-extension';
 import { SuggestExtension, SuggestOptions } from './suggest-extension';
 import { TagsExtension } from './tags-extension';
 
-export interface BuiltinOptions extends SuggestOptions, KeymapOptions {}
+export interface BuiltinOptions extends SuggestOptions, KeymapOptions, PersistentSelectionOptions {}
 
 /**
  * Provides all the builtin extensions to the editor.
@@ -78,6 +82,7 @@ export class BuiltinPreset extends Preset<BuiltinOptions> {
       'selectParentNodeOnEscape',
       'undoInputRuleOnBackspace',
     ]);
+    const persistentSelectionOptions = pick(this.options, ['persistentSelectionClass']);
 
     return [
       // The order of these extension is important.
@@ -92,6 +97,7 @@ export class BuiltinPreset extends Preset<BuiltinOptions> {
       new CommandsExtension(),
       new HelpersExtension(),
       new KeymapExtension(keymapOptions),
+      new PersistentSelectionExtension(persistentSelectionOptions),
     ];
   }
 }
--- a/packages/@remirror/core/src/builtins/index.ts
+++ b/packages/@remirror/core/src/builtins/index.ts
@@ -7,6 +7,7 @@ export * from './keymap-extension';
 export * from './node-views-extension';
 export * from './paste-rules-extension';
 export * from './plugins-extension';
+export * from './persistent-selection-extension';
 export * from './schema-extension';
 export * from './suggest-extension';
 export * from './tags-extension';
--- a/packages/@remirror/extension-annotation/src/__tests__/annotation-extension.spec.ts
+++ b/packages/@remirror/extension-annotation/src/__tests__/annotation-extension.spec.ts
@@ -22,14 +22,14 @@ describe('commands', () => {
     commands.addAnnotation({ id: '1' });
 
     expect(view.dom.innerHTML).toMatchInlineSnapshot(`
-          <p>
-            An
-            <span class="annotation">
-              important
-            </span>
-            note
-          </p>
-        `);
+      <p>
+        An
+        <span class="selection annotation">
+          important
+        </span>
+        note
+      </p>
+    `);
   });
 
   it('#updateAnnotation', () => {
@@ -40,28 +40,28 @@ describe('commands', () => {
 
     // Pre-condition
     expect(view.dom.innerHTML).toMatchInlineSnapshot(`
-          <p>
-            An
-            <span class="annotation">
-              important
-            </span>
-            note
-          </p>
-        `);
+      <p>
+        An
+        <span class="selection annotation">
+          important
+        </span>
+        note
+      </p>
+    `);
 
     commands.updateAnnotation(id, {
       className: 'updated',
     });
 
     expect(view.dom.innerHTML).toMatchInlineSnapshot(`
-          <p>
-            An
-            <span class="annotation updated">
-              important
-            </span>
-            note
-          </p>
-        `);
+      <p>
+        An
+        <span class="selection annotation updated">
+          important
+        </span>
+        note
+      </p>
+    `);
   });
 
   it('#setAnnotations', () => {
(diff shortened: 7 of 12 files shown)


ankon:
Two open questions:

1. Documentation: Sure, where? :)
2. Changeset: Uuuh. Do I do that, or do you do that? I suspect it involves running `pnpm changeset`.

ankon:
> Changeset: Uuuh. Do I do that, or do you do that? I suspect it involves running pnpm changeset.

Right, I think I understand -- changeset == documentation piece. I added something, and I think this is a "patch" change. If the user doesn't configure the extension, then nothing will change.

.changeset/unlucky-pets-melt.md
```diff
@@ -0,0 +1,9 @@
+---
+'@remirror/core': patch
```
ifiokjr, line 2: Since this is a new feature, make it a minor change.

```suggestion
'@remirror/core': minor
```

packages/@remirror/core/src/builtins/__tests__/persistent-selection-extension.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    view: { props },
+    state,
+    nodes: { p, doc },
+  } = renderEditor([], { builtin });
+
+  add(doc(p('Some text')));
+
+  selectText(4);
+
+  setTimeout(() => {
```
ifiokjr, line 21: Not sure that we need the `setTimeout` here. Updates are immediately available. 

Also, the test seems to be focused on the implementation details. IMO the important things to test are:

- The current selection is wrapped with an element that has the class `persistent-selection`. So `expect(view.dom.innerHTML).toMatchInlineSnapshot()` is a good enough test.
- When focus is lost the selection class is maintained. I think this was the main goal of the original issue.

packages/@remirror/core/src/builtins/__tests__/persistent-selection-extension.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    add,
+    selectText,
+    view: { props },
+    state,
+    nodes: { p, doc },
+  } = renderEditor([], { builtin });
+
+  add(doc(p('Some text')));
+
+  selectText(4);
```
ifiokjr, line 19: Selection should probably be a range here. `selectText(4, 9)`. It raises a good question. Do we want the selection to even be wrapped when empty?

ifiokjr: @ronnyroeller, any thoughts?

packages/@remirror/core/src/builtins/persistent-selection-extension.ts
```diff
@@ -0,0 +1,50 @@
+// XXX: @remirror/core exports a _type_ 'DecorationSet', but we need access to the static methods as well.
```
ifiokjr, line 1: I'm not sure what this comment means. Do you want to export the `DecorationSet` from `@remirror/core`?

packages/@remirror/core/src/builtins/persistent-selection-extension.ts
```diff
@@ -0,0 +1,50 @@
...
+import { ExtensionPriority } from '@remirror/core-constants';
+import type { Static } from '@remirror/core-types';
+import { Decoration, DecorationSet } from '@remirror/pm/view';
+
+import { extensionDecorator } from '../decorators';
+import { PlainExtension } from '../extension';
+import type { CreatePluginReturn } from '../types';
+
+export interface PersistentSelectionOptions {
+  persistentSelectionClass?: Static<string>;
```
ifiokjr, line 12: Good use of `Static` 👍

packages/@remirror/core/src/builtins/persistent-selection-extension.ts
```diff
@@ -0,0 +1,50 @@
...
+    return {
+      props: {
+        decorations: ({ doc, selection }) => {
+          // Do not run the extension at all when there is no actual decoration
+          // to be done.
+          if (!this.options.persistentSelectionClass) {
+            return DecorationSet.empty;
+          }
+
+          if (selection.empty) {
```
ifiokjr, line 37: Now I'm not even sure how the previous test passes 😄 
It looks like you return early for empty selections.

ifiokjr: Oh I see, it passes because it's never run. The `setTimeout` means that the error will be thrown after the test.

ifiokjr (changes requested):

ifiokjr:
Ooops I submitted the review too early last time without comment 😢  

Here's the reason for requesting changes.

> The editor should give the illusion that the current selection stays even when the editor lost focus

I think this part of issue #601 hasn't really been resolved. There is a selection added to the editor, however, it's not been related to the whether the editor is  in focus. Unfortunately, docs don't build for PR's from an outside fork, so I can't try out your code in the playground without building it myself. 

I've added @ronnyroeller as a reviewer as well.

packages/@remirror/core/src/builtins/persistent-selection-extension.ts
```diff
@@ -0,0 +1,50 @@
...
+import { PlainExtension } from '../extension';
+import type { CreatePluginReturn } from '../types';
+
+export interface PersistentSelectionOptions {
+  persistentSelectionClass?: Static<string>;
+}
+
+@extensionDecorator<PersistentSelectionOptions>({
+  defaultOptions: {
+    persistentSelectionClass: undefined,
```
ifiokjr, line 17: I actually feel like there should a default value here. Something like `selected-text`.

.changeset/unlucky-pets-melt.md
```diff
@@ -0,0 +1,9 @@
...
+---
+
+Optionally allow to style the currently selected text
```
ifiokjr, line 5: Nice use of changesets here 👍

packages/@remirror/core/src/builtins/persistent-selection-extension.ts
ankon: I initially had a default, but that will break compatibility (as many of the unit tests show). 

Of course, the question is whether 100% compatibility is relevant here in the context of remirror.

packages/@remirror/core/src/builtins/__tests__/persistent-selection-extension.spec.ts
ankon: I had to add the timeout, neither `act()` nor simply chaining the calls passed the tests.

packages/@remirror/core/src/builtins/__tests__/persistent-selection-extension.spec.ts
ankon: `selectText` defaults the end to the end of the text, but I can make this explicit of course.

packages/@remirror/core/src/builtins/persistent-selection-extension.ts
ifiokjr: Oh, that's a valid point. 

I don't mind updating the snapshots, though. It's a core piece of functionality and my feeling is that it will be an expected behaviour for the majority of users. Also, the class is pretty harmless by itself, unless styles are assigned.

packages/@remirror/core/src/builtins/persistent-selection-extension.ts
ankon: `@remirror/Core` exports the `DecorationSet` and `Decoration` types, but using `DecorationSet.empty` lead to Compiler errors as that would need the static property. 

Arguably: I can remove the comment if importing from `@remirror/pm/view` is ok.

ankon:
As to addressing #601: what I found is that the selection doesn't get lost at all, the problem is merely a display issue of not being able to _see_ the selection. 

For that reason this extension/plugin is fairly trivial, it just needs to keep a decoration on the selected area: as soon as the browser removes whatever it does by default the decoration takes over, thereby keeping things visible.

ankon:
Moved it back to draft -- I think I see the issue with the tests, so I'll start from there and then update the branch when I have tests that actually test the right thing.

packages/@remirror/core/src/builtins/persistent-selection-extension.ts
ifiokjr: Yeah, importing from `@remirror/pm/view` is expected. The types exported by `@remirror/core-types` are all aliases. Initially, it was to reduce the number of import statements needed when creating an extension. The types for `EditorView`, `EditorSchema`, `Selection`, `ProsemirrorNode`, `Mark` are used so often that it made sense to alias them.

ankon:
Alright, next update, addressing most of the feedback.

> Oh I see, it passes because it's never run. The `setTimeout` means that the error will be thrown after the test.

Indeed, the tests actually didn't do anything sensible. The problem here was actually that looking at the `props.decorations` was pointless (it contains incoming properties, and not a view on the _current_ properties as I wrongly assumed). The new tests simply use Jest's snapshots, which works just fine to show the extension getting activated.

I still need to look into writing tests that show that a change in focus away from the component still leaves these decorations around.

One big issue though is the second commit in this set, where I don't really know the right direction forward. The code-block extension creates simple `<span>text</span>` decorations for `text`-typed refractor nodes, and when the persistent-selection extension adds a decoration on exactly one of these we end up with a calcuated `class` of "persistent-selection " (note the trailing space). prosemirror-view then tries to parse that value into tokens, so that it can maintain the `classList` DOMTokenList manually: https://github.com/ProseMirror/prosemirror-view/blob/b9fd1bc695a9c69ac3c472ffdbd61115b51d0588/src/viewdesc.js#L962
The problem in prosemirror-view is that it may produce empty tokens, which are explicitly illegal in calls to `DOMTokenList.add` or `DOMTokenList.remove` (see https://dom.spec.whatwg.org/#dom-domtokenlist-add). I'm thinking that prosemirror-view is to blame here, and that it should deduplicate/remove excess whitespace (as MDN hints here: https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList#Trimming_of_whitespace_and_removal_of_duplicates). But at the same time it also looks weird to produce decorations without a class, so I added a 'text' class in the code-block extension, which _also_ fixes the problem by avoiding the prosemirror-view issue. WDYT?

ifiokjr:
Okay, thanks for the write-up, although I'm not sure I fully understand. Is there an error being produced?

ankon:
The error is really looking innocent enough when you only apply the first commit:
~~~~
  ● plugin › Backspace › avoids stepping into the previous node with non empty selection

    SyntaxError: The token provided must not be empty.
~~~~

The call stack of that error is hidden, and looks like this:
~~~~
validateTokens (.../remirror/node_modules/.pnpm/jsdom@16.4.0/node_modules/jsdom/lib/jsdom/living/nodes/DOMTokenList-impl.js:13)
add (.../remirror/node_modules/.pnpm/jsdom@16.4.0/node_modules/jsdom/lib/jsdom/living/nodes/DOMTokenList-impl.js:110)
add (.../remirror/node_modules/.pnpm/jsdom@16.4.0/node_modules/jsdom/lib/jsdom/living/generated/DOMTokenList.js:159)
patchAttributes (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:967)
patchOuterDeco (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:949)
applyOuterDeco (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:981)
create (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:597)
addNode (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:1131)
(anonymous function) (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:656)
iterDeco (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:1249)
updateChildren (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:639)
updateInner (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:725)
update (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:717)
updateNextNode (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:1116)
(anonymous function) (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:654)
iterDeco (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:1197)
updateChildren (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:639)
updateInner (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:725)
update (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/viewdesc.js:717)
updateStateInner (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/index.js:160)
updateState (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/index.js:114)
updateState (.../remirror/packages/@remirror/dom/src/dom.ts:114)
(anonymous function) (.../remirror/packages/@remirror/core/src/editor-wrapper.ts:340)
dispatch (.../remirror/node_modules/.pnpm/prosemirror-view@1.15.5/node_modules/prosemirror-view/src/index.js:367)
dispatchTextSelection (.../remirror/packages/jest-prosemirror/src/jest-prosemirror-editor.ts:188)
(anonymous function) (.../remirror/packages/jest-remirror/src/jest-remirror-editor.ts:309)
(anonymous function) (.../remirror/packages/@remirror/extension-code-block/src/__tests__/code-block-extension.spec.ts:147)
Promise.then.completed (.../remirror/node_modules/.pnpm/jest-circus@26.4.2/node_modules/jest-circus/build/utils.js:230)
callAsyncCircusFn (.../remirror/node_modules/.pnpm/jest-circus@26.4.2/node_modules/jest-circus/build/utils.js:170)
_callCircusTest (.../remirror/node_modules/.pnpm/jest-circus@26.4.2/node_modules/jest-circus/build/run.js:212)
[ async function ]
_callCircusTest (.../remirror/node_modules/.pnpm/jest-circus@26.4.2/node_modules/jest-circus/build/run.js:197)
_runTest (.../remirror/node_modules/.pnpm/jest-circus@26.4.2/node_modules/jest-circus/build/run.js:149)
[ async function ]
_runTest (.../remirror/node_modules/.pnpm/jest-circus@26.4.2/node_modules/jest-circus/build/run.js:106)
_runTestsForDescribeBlock (.../remirror/node_modules/.pnpm/jest-circus@26.4.2/node_modules/jest-circus/build/run.js:63)
[ async function ]
_runTestsForDescribeBlock (.../remirror/node_modules/.pnpm/jest-circus@26.4.2/node_modules/jest-circus/build/run.js:36)
_runTestsForDescribeBlock (.../remirror/node_modules/.pnpm/jest-circus@26.4.2/node_modules/jest-circus/build/run.js:57)
[ async function ]
_runTestsForDescribeBlock (.../remirror/node_modules/.pnpm/jest-circus@26.4.2/node_modules/jest-circus/build/run.js:36)
_runTestsForDescribeBlock (.../remirror/node_modules/.pnpm/jest-circus@26.4.2/node_modules/jest-circus/build/run.js:57)
~~~~

The problem occurs in the prosemirror-view `ViewDesc.create` call, where it needs to apply 2 decorations:
```json
[
  {
    "from": 0,
    "to": 4,
    "type": {
      "spec": {},
      "attrs": {
        "class": "persistent-selection"
      }
    }
  },
  {
    "from": 0,
    "to": 4,
    "type": {
      "spec": {},
      "attrs": {
        "class": ""
      }
    }
  }
]
```

The `patchAttributes` call is trying to update the `class` attribute with the value "persistent-state " (trailing space!), when then fails when it tries to add the "new" empty token to the `classList`.

ankon:
The TL;DR of the question is probably: Are you ok with another PR that would change the code-block extension to set a (configurable) class on `text` nodes, and that would otherwise not create annotations for `text` nodes?

ifiokjr:
> The TL;DR of the question is probably: Are you ok with another PR that would change the code-block extension to set a (configurable) class on `text` nodes, and that would otherwise not create annotations for `text` nodes?

I'm okay with that. Meanwhile, we could just ignore the code block, via code. If the parent node is a code block, don't create a decoration, if that's easier.

ankon:
> I'm okay with that. Meanwhile, we could just ignore the code block, via code. If the parent node is a code block, don't create a decoration, if that's easier.

I did quite some back-and-forth here, but ultimately I think the approach of one the one hand preventing to create decorations without a `class`, and on the hand still retaining the ability to have all parts of a code block targetable through CSS, is the best way forward here. I'm will still create an issue for prosemirror-view as well, as the situation seems somewhat "improveable" :)

See https://github.com/remirror/remirror/pull/630.

**For now I kept this PR on top of #630, as otherwise the tests will fail. I will rebase onto the `next` branch before marking this "ready for review"**

ifiokjr:
Once you've updated this branch, I'll merge in your work.

packages/@remirror/core/src/builtins/__tests__/persistent-selection-extension.spec.ts
ankon: > When focus is lost the selection class is maintained.

Added a test for this case as well now.

ankon:
Thank you for your patience! I think from my side I covered now everything from the review comments as well, so "should be good to go" :)

packages/@remirror/core/src/builtins/__tests__/persistent-selection-extension.spec.ts
```diff
@@ -0,0 +1,60 @@
+import { extensionValidityTest, renderEditor } from 'jest-remirror';
+
+import { PersistentSelectionExtension } from '..';
+
+extensionValidityTest(PersistentSelectionExtension);
+
+describe('configuration', () => {
+  [
```
ifiokjr, line 8: Jest actually has it's own builtin `.each` test method. 

https://jestjs.io/docs/en/api#testeachtablename-fn-timeout

packages/@remirror/core/src/builtins/__tests__/persistent-selection-extension.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+      add(doc(p('Some text')));
+      selectText({ from: 4, to: 9 });
+
+      expect(dom.outerHTML).toMatchSnapshot();
+    });
+  });
+});
+
+describe('focus', () => {
+  test('retains selection and decoration', () => {
```
ifiokjr, line 31: Nitpick, I prefer `test()` when it's at top level and `it()` when nested in a `describe` block. 

I don't even know why, and you can leave it as is.

packages/@remirror/core/src/builtins/__tests__/persistent-selection-extension.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    add(doc(p('Some text')));
+    const range = { from: 4, to: 9 };
+    selectText(range);
+
+    // Verify that the selection has been applied
+    expect(editor.state.selection.from).toBe(range.from);
+    expect(editor.state.selection.to).toBe(range.to);
+
+    // Add a new element and focus that one
```
ifiokjr, line 49: You can also use `editor.view.dom.blur()`.

.changeset/unlucky-pets-melt.md
```diff
@@ -0,0 +1,9 @@
+---
+'@remirror/core': minor
+---
+
+Optionally allow to style the currently selected text
+
+This adds a new option for the builtin preset, `persistentSelectionClass`. If that is set to a valid CSS class name any selection in the editor will be decorated with this class.
+
+This can be used to keep a indication for the current selection even when the focus changes away from the editor.
```
ifiokjr, line 9:
```suggestion
This can be used to keep an indication for the current selection even when the focus changes away from the editor.
```

ifiokjr (approved):
The tests are looking much better 🥳 

Just some super, super minor tweaks which you can feel free to ignore. I'll merge in a few hours regardless.

packages/@remirror/core/src/builtins/__tests__/persistent-selection-extension.spec.ts
ankon: I am starting to like jest :)

packages/@remirror/core/src/builtins/__tests__/persistent-selection-extension.spec.ts
ankon: I needed to explicitly cast the `dom` property to `HTMLElement` to have access to `blur()`, but this definitely reads better and makes it more obvious what the point of the test is.

ifiokjr: Oh interesting, try `editor.dom`, just realised that might already be cast to an HTML element.

ifiokjr: Ignore that, it isn't. I'll update it so that it is.

PR #617, opened 2020-09-01
Update inline documentation for `createPlugin`

ankon:
### Description

Updated documentation based on reading code and better understanding of the programming model.

### Checklist

- [x] ~New code is unit tested and all current tests pass when running `pnpm test`.~ (no new code)

### DIFF
--- a/packages/@remirror/core/src/builtins/plugins-extension.ts
+++ b/packages/@remirror/core/src/builtins/plugins-extension.ts
@@ -377,7 +377,7 @@ declare global {
        * the plugin state.
        *
        * ```ts
-       * import { CreatePluginReturn } from '@remirror/core';
+       * import { CreatePluginReturn } from 'remirror/core';
        *
        * class MyExtension extends PlainExtension {
        *   get name() {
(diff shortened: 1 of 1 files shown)


packages/@remirror/core/src/builtins/plugins-extension.ts
```diff
@@ -377,15 +377,15 @@ declare global {
        * the plugin state.
        *
        * ```ts
-       * import { ExtensionPluginSpec } from 'remirror/core';
+       * import { CreatePluginReturn } from '@remirror/core';
```
ifiokjr, line 380: Good spot!

ifiokjr: Although it should be `remirror/core`. It's much easier for users to install `remirror` with everything included than install each package individually. Since it's the recommended way of getting started I've decided for all the core docs to assume that's the way it's being used. Advanced users will decide which they prefer and can probably figure out the difference for themselves.


```suggestion
       * import { CreatePluginReturn } from 'remirror/core';
```

ifiokjr (changes requested):
Thanks for this. Looks good, just one change to make before merging.

ifiokjr (approved):

ifiokjr:
Looks great 👍

PR #618, opened 2020-09-01
Update changeset config to point to `next`

ankon:
### Description

Set the base branch to the current branch name, `next`.

### Checklist

- [x] ~I have updated the documentation where necessary.~
- [x] ~New code is unit tested and all current tests pass when running `pnpm test`.~ (no new code)

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -84,5 +84,5 @@
     ]
   ],
   "access": "public",
-  "baseBranch": "main"
+  "baseBranch": "next"
 }


.changeset/config.json
```diff
@@ -84,5 +84,5 @@
     ]
   ],
   "access": "public",
-  "baseBranch": "main"
+  "baseBranch": "next"
```
ifiokjr, line 87: Oh cool, I assume that this is because `pnpm changeset` doesn't work unless the base branch exists in your local branches. Valid. 

I think it actually  makes more sense to just remove this line. I just had a look at how [`emotion`](https://github.com/emotion-js/emotion/blob/master/.changeset/config.json#L1-L30) which is by the creator of `changesets` does things, and they don't seem to have it at all. 

My concern is that in the current prerelease mode, it'll remove all the active changesets if the current branch is seen to be the main one. It probably won't, though.

ankon: I don't think it'll "forget" the changes, these are after all just a "diff" between what's in this directory and what's stored inside `.changesets/pre.json`. 

But: [The default for this value seems to be `master`](https://github.com/atlassian/changesets/blob/f9fd2835aa86fed1c3e7df2613a7550e33f9cc2e/packages/config/src/index.ts#L248-L251), which might be as worse if not worse than the current default: `main` doesn't exist, so you just get an error -- `master` does exist, and is horribly old :)

ifiokjr: Okay, makes sense.

Thanks for testing 😄 

In that case I'm happy to merge. If it causes issues I'll be able to see them in the changesets PR.

PR #628, opened 2020-09-03
Make the storybook package start up

ankon:
### Description

* Explicitly switch ts-node into commonjs mode, which seems to be required by
  the specific configuration of storybook here
* Disable an eslint issue about not having default exports on the main
  storybook module

### Checklist

- [x] ~New code is unit tested and all current tests pass when running `pnpm test`.~: No new code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "reset": "pnpm clean:all; rm pnpm-lock.yaml; pnpm symlink:root; pnpm install;",
     "run:prettier": "prettier --ignore-path .eslintignore \"**/*.{js,jsx,ts,tsx,md,mdx,json,html,css,yml,yaml,graphql}\"",
     "size": "size-limit",
-    "storybook": "manypkg run support/storybook storybook",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -68,7 +68,7 @@
     "reset": "pnpm clean:all; rm pnpm-lock.yaml; pnpm symlink:root; pnpm install;",
     "run:prettier": "prettier --ignore-path .eslintignore \"**/*.{js,jsx,ts,tsx,md,mdx,json,html,css,yml,yaml,graphql}\"",
     "size": "size-limit",
-    "storybook": "manypkg run support/storybook storybook",
+    "storybook": "cross-env TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\"}' manypkg run support/storybook storybook",
```
ifiokjr, line 71: I wasn't aware of this. Has it changed recently?

ifiokjr (approved):
Looks good

package.json
ankon: TBH: I really don't know. Not providing this produces errors from babel, so it could also be that a babel default has changed ... or maybe something in react-scripts/webpack? This was a quick-and-easy way to fix things up, and is nice and localized to not disturb anything else. :)

Issue #647, opened 2020-09-05
Attempted import error: 'en' is not exported from 'make-plural/plurals'.


anotheri:
### Summary

I've opened ticket as [we discussed](https://github.com/remirror/remirror/discussions/643).

I'm Using webpack 4 and all the latest remirror packages:

```
"remirror": "^1.0.0-next.32",
"@remirror/pm": "^1.0.0-next.32",
"@remirror/i18n": "^1.0.0-next.32",
```

And I'm getting this error on build:

```
Attempted import error: 'en' is not exported from 'make-plural/plurals'.
```

My `@ ./src/shared/remirrorProvider.js` file looks exactly like one of your basic examples
```
import React from 'react';
import { SocialEmojiComponent, SocialProvider } from 'remirror/react/social';

export const Remirror = () => {
  return (
    <SocialProvider>
      <SocialEmojiComponent />
    </SocialProvider>
  );
};
```

### Expected results

Build pass.

### Actual results

```
Creating an optimized production build...
(node:56701) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead
^[Failed to compile.

./node_modules/@remirror/i18n/dist/i18n.browser.esm.js
Attempted import error: 'en' is not exported from 'make-plural/plurals'.
 @ ./node_modules/@remirror/react/dist/react.browser.esm.js
 @ ./node_modules/remirror/react/dist/remirror.browser.esm.js
 @ ./src/shared/remirrorProvider.js
```

ifiokjr:
In the lockfile, what version of `make-plural` is installed. 

I'm looking at [unpkg](https://unpkg.com/browse/make-plural@6.2.2/plurals.js) and it's right there on [line: 213](https://unpkg.com/browse/make-plural@6.2.2/plurals.js) and in the [`mjs` file (line: 166)](https://unpkg.com/browse/make-plural@6.2.2/plurals.mjs) as well. 


I don't know whether webpack 4 is able to resolve the `exports` field yet. So it might have to do with the fact that they're using it to define the main entry points in the [`package.json`](https://unpkg.com/browse/make-plural@6.2.2/package.json) file.

But that doesn't make sense. They also have a `main`, `browser` and `module` field.

anotheri:
```
yarn why v1.22.4
[1/4] 🤔  Why do we have the module "make-plural"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[...]

PR #690, opened 2020-09-16
Update errors.md

anotheri:
Update documentation as it has been discussed in this issue: https://github.com/remirror/remirror/issues/647

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -312,35 +312,39 @@ There is something wrong with your i18n setup.
 > Attempted import error: 'en' is not exported from 'make-plural/plurals'.
 
 There is a problem related to Webpack (4 & 5) build configuration. Make sure:
-- `file-loader` does not process `.mjs` files, e.g.:
-```
-// "file" loader makes sure assets end up in the `build` folder.
-// When you `import` an asset, you get its filename.
-// This loader doesn't use a "test" so it will catch all modules
-// that fall through the other loaders.
-{
-  loader: require.resolve('file-loader'),
 
-  // Exclude `js` files to keep "css" loader working as it injects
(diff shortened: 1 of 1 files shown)


docs/errors.md
```diff
@@ -304,3 +304,41 @@ Something went wrong with your custom ReactNodeView Component.
 > I18N Context
 
 There is something wrong with your i18n setup.
+
+### RMR0301
```
ifiokjr, line 308: Could you change the title to the following? It's not actually possible for the **remirror* runtime to capture the custom error `RMR0301` since it happens during the build process. 

```suggestion
## Build Errors

### `make-plural/plurals`
```

ifiokjr (changes requested):
Appreciate you taking the time out to create this PR 🙌 

Just one change that I've noticed. 

Also, you're probably going to need to run `pnpm fix` since the markdown linting is likely to fail.

anotheri:
@ifiokjr that lint fail has 228 warnings, I fixed just readme and all my PR has less lines :)
do you suggest me to fix all those warnings or what?

ifiokjr:
If you run `pnpm fix` it should fix the build.

The `ESLint` warnings aren't relevant. I'll make sure to turn them off for CI builds in a separate PR since as you correctly point out, they are verbose and confusing.

The failure is actually from a prettier failure in the markdown file.

anotheri:
@ifiokjr hey, I've fixed formatting with `pnpm fix`, not sure why it fails now

ifiokjr (approved):
> I've fixed formatting with `pnpm fix`, not sure why it fails now

I think that's due to some flakiness in the Azure CI. I'm happy to merge as it is.

PR #706, opened 2020-09-24
feat(v1): prepare for `@1` release

### LINKED ISSUE #37: Add new package `@remirror/extension-native-bridge`
## Description

The goal of this project is to be an editor available on all platforms. Handling text in a WebView with bidirectional communication via React Native, Expo helps a lot in this regard.

`Flutter` and `NativeScript` could be looked at post 1.0.

## Possible Implementation

- Create a new package called `@remirror/native-bridge`
  - This extension will be responsible for handling bi-directional communication. It will take in events from the native layer via `window.onMessage` and post messages to the native layer via `window.postMessage`. 

**[STATUS UPDATE](https://github.com/remirror/remirror/issues/37#issuecomment-688686546)**

I'm looking for two or three sponsors who'd like to champion this feature. If you're not in a position to sponsor but would still like to see this integrated into **remirror**, please reply with your use cases and upvote this comment for more visibility.

It's something I really want to work on, but am also aware it's a big undertaking.

### LINKED ISSUE #407: Improve playground loading screen and styles
## Description

@benjie has been doing really cool work on the [playground](http://localhost:3001/playground). It's got to the point where we've started to care about how it looks. 

This task is for anyone who enjoys that kind of thing. 

- [ ] Style the loading screen.
- [ ] Fix the width on larger screens (white edges).
- [ ] _OPTIONAL_ - Give the rendered editor better default styles.

I'm available to help anyone who has any questions. Feel free to reach out on the [discord channel](https://discord.gg/C4cfrMK). 

## Checklist

### LINKED ISSUE #453: Create floating menu components
### Description

Floating menu components include.

- Bubble menu on selection (with positioner API)
- Bubble menu on selection (with popperjs)
- Inline menu (for showing menu at the start of the line).

### LINKED ISSUE #480: Enable excluding `Extension` by name or constructor from `Preset`
### Description

Similar to the default extension options, `Preset`'s should have a default option (available to all presets) for excluding extensions by name or type.

This builds on the work done in #415

### LINKED ISSUE #504: Rewrite `@remirror/extension-emoji` to support emoji nodes
### Description

Currently, emoji is saved as text only. This causes issue on machines without support for the full emoji specification. It also means the rendering is limited to how your operating systems renders emoji. 

It would be nice to have greater control. 

- [ ] Add support for rendering emoji as nodes
- [ ] Adapt search from [emoji-mart](https://github.com/missive/emoji-mart/blob/master/src/utils/emoji-index/nimble-emoji-index.js).
- [ ] Extract emoji data to a separate package (or external package).
-  [ ] Create SVG URL object structure for the following libraries:
 - [noto-emoji](https://github.com/missive/emoji-mart/blob/master/src/utils/emoji-index/nimble-emoji-index.js)
 - [openmoji](https://github.com/hfg-gmuend/openmoji) 
 - [blobmoji](https://github.com/c1710/blobmoji)
 - [twitter emoji](https://github.com/twitter/twemoji) - might need attribution
 - [github emoji](https://github.com/seoyoochan/github-emoji-svg) - might be outdated
 - apple emoji if SVG's can be extracted [here](https://github.com/iamcal/emoji-data#building-the-apple-emoji-images)
 - [emojidex](https://github.com/emojidex/emojidex-vectors) - if the quality is high enough
 - [emojitwo](https://github.com/EmojiTwo/emojitwo) - make sure the license supports this use case.
The prosemirror representation of the emoji will be as follows:

```json
{
  "type": "emoji",
  "attrs": {
    "shortName": ":smile:",
    "id": "1f604",
    "name": "at",
    "skinVariation": 0,
    "native": "😄" 
  }
}
```

### Context

One of the reasons I moved away from using emoji nodes in the previous version was due to poor support on mobile devices. Android currently can't handle atom nodes without losing its position constantly. I spent a lot of time trying workarounds, but it's a real issue. 

Discord, which uses `SlateJS`, works around this by rendering native emoji on mobile devices. 

- [ ] Add `isMobile` property to the `Extension.store` which is available in every extension. The end user can set the value to true or false. By default, this will be undefined.

A potential bug on some versions of Chrome when rendering native emoji was mentioned in #448 . There seems to be overlap between the text and the emoji in certain situations.

- [ ] Fix emoji / text overlap on Chrome in rewrite.

### LINKED ISSUE #508: Create a `Migration Guide` for users moving from `0.x` to `remirror@1.0.0-next`
### Description

Based on this [comment](https://github.com/remirror/remirror/issues/255#issuecomment-657285855) and this [discussion](https://github.com/remirror/remirror/discussions/357) I need to create a migration guide for users coming from the older version of `remirror`.

- [ ] Add a migration section to the documentation website https://remirror.io/documentation/migration/0-to-1
- [ ] Update the discussion and issue mentioned above.

This issue is closely linked to #476.

### LINKED ISSUE #531: New package `@remirror/extension-columns`
### Description

Mentioned during a discussion with @bertg was the idea of adding support for multi column nodes. It wasn't on the roadmap at the time, but it turns out the implementation is not so difficult. 

- [x] Create `ColumnsExtension` which is a block node that wraps other block nodes. It can include paragraphs, headings etc.
- [x] It adds an extra attribute to all block nodes `columnSpan`.

### Context

The following properties should be supported.

| Property            | Description                                                        |
| ------------------- | ------------------------------------------------------------------ |
| `column-count`      | Specifies the number of columns an element should be divided into  |
| `column-fill`       | Specifies how to fill columns                                      |
| `column-gap`        | Specifies the gap between the columns                              |
| `column-rule`       | A shorthand property for setting all the column-rule-\* properties |
| `column-rule-color` | Specifies the color of the rule between columns                    |
| `column-rule-style` | Specifies the style of the rule between columns                    |
| `column-rule-width` | Specifies the width of the rule between columns                    |
| `column-span`       | Specifies how many columns an element should span across and used by child blocks to span multiple columns           |
| `column-width`      | Specifies a suggested, optimal width for the columns               |
| `columns`           | A shorthand property for setting column-width and column-count     |

### LINKED ISSUE #532: New package `@remirror/preset-formatting`
### Description

Create a preset with all the formatting node and mark extensions included.

Integrate the mark extensions which impact formatting.

- [ ] `BoldExtension`
- [ ] `UnderlineExtension`
- [ ] `ItalicExtension`
- [ ] `StrikeExtension`
- [ ] `CodeExtension` 🤷 
- [ ] _`ColorExtension`_ #535
- [ ] _`HighlightExtension`_ #536
- [ ] _`FontSizeExtension`_ #537
- [x] _`FontFamilyExtension`_ #538
- [x] _`SupExtension`_ #539
- [x] _`SubExtension`_ #540
- [x] _`CaseExtension`_ #541

Integrate the node extensions which impact formatting.

- [ ] _`ColumnExtension`_ #531
- [ ] _`ContentBlockExtension`_ #542
- [x] _`NodeFormattingExtension`_ #530

Other updates.

- [ ] Enable override of default tag to use for marks that support it.

### LINKED ISSUE #533: Add TypeScript autocompletion and intellisense to playground
### Description

- [ ] Add full support for TypeScript intellisense to all local packages.
- [ ] Explore adding support for external packages.

### LINKED ISSUE #535: New package `@remirror/extension-text-color`
### Description

Wraps text with a styled `span` using the `color` css property. The name of the wrapper tag should be configurable.

- [ ] Create `TextColorExtension`

### LINKED ISSUE #536: New package `@remirror/extension-text-highlight`
### Description

Add a color background to the selected text (or text within a specified range).

- [ ] Create `TextHighlightExtension`

### LINKED ISSUE #537: New package `@remirror/extension-font-size`
### Description

Add font size to the selected text, or text within the provided range.

- [x] Create `FontSizeExtension`

### LINKED ISSUE #538: New package `@remirror/extension-font-family`
### Description

Add a font family to selected text, or text within the provided range.

- [ ] Create `FontFamilyExtension`

### LINKED ISSUE #541: New package `@remirror/extension-text-case`
### Description

Wraps text in a `span` with support `small-caps` via the `font-variant` style and `none` (default), `upper`, `lower` and `capitalize` via the `text-transform` style.

- [x] Create `CaseExtension`

### LINKED ISSUE #542: New package `@remirror/extension-content-block`
### Description

A block node that holds other block nodes. This enables a block based style of editing where the editor is split into separate blocks of functionality.

- [ ] Create `ContentBlockExtension`

### LINKED ISSUE #566: Add react-popper component to `@remirror/react-components`
Add a component which can be used in the playground for positioning using `react-popper`.

- [ ] This should use the positioner API to create [`@popperjs/core@2.x` virtual elements](https://popper.js.org/docs/v2/virtual-elements/)

### LINKED ISSUE #569: Refactor playground
### Description

General refactoring.

### LINKED ISSUE #652: Create Twitch account for `remirror`
### Description

One day, I'd like to stream something.

### LINKED ISSUE #695: Improve API discoverability
### Description

Right now it's hard to know where to import things from. Make the API more discoverable by making everything available via the top level `remirror` import.

- [x] Export all core exports from the `remirror` entry point.
- [x] Export all presets and extensions from the `remirror/extensions` entry point.
- [x] Export all react stuff from `remirror/react` with no nesting.
- [x] Export all dom packages from `remirror/dom`

### LINKED ISSUE #709: Add package `@remirror/extension-text-wrap`
### Description

Add extension which provides the ability to wrap text when certain characters are entered.

The configuration will only be active for text selections.

```ts
import type { CommandFunction, EditorState } from '@remirror/core';
import type { TextSelection } from '@remirror/pm/state';

export interface TextWrap {
  /** 
   * The unique identifier for this text wrap configuration. This is to support
   * the use case where multiple text wraps with identical open and close
   * characters are added but with different criteria and options. You might
   * want different behaviour when active in a heading node as opposed to a
   * paragraph.
   */
  name: string;

  /**
   * The opening character. Single characters only.
   */
  open: string;

  /**
   * The closing character. Single characters only.
   */
  close: string;

  /**
   * Valid characters contained by the selection
   */
  regex?: RegExp | string;

  /**
   * Criteria to meet before wrapping.
   *
   * @default TextWrapCriteria.Always
   */
  criteria?: TextWrapCriteria | TextWrapCriteriaFunction;

  /**
   * What to do on backwards deletions. This basically intercepts the
   * `Backspace` keyboard handler.
   *
   * @default TextWrapDeletions.ContentEmpty
   */
  backwardDelete?: TextWrapDeletions | CommandFunction;

  /**
   * What to do on forward deletions. This basically intercepts the `Delete`
   * keyboard handler.
   *
   * @default TextWrapDeletions.ContentEmpty
   */
  forwardDelete?: TextWrapDeletions | CommandFunction;

  /**
   * A list of node names and groups which are not supported for this rule.
   *
   * @default ['code']
   */
  invalidNodes?: string[];

  /**
   * A list of supported node names and groups. This takes priority over
   * `invalidNodes` if both properties are present.
   *
   * Setting this to an empty array would mean that this rule would never be
   * matched.
   *
   * @default null
   */
  validNodes?: string[] | null;

  /**
   * A list of mark names and groups which will invalidate the text wrap.
   *
   * @default ['code']
   */
  invalidMarks?: string[];

  /**
   * A list of supported mark names. This takes priority over `invalidMarks` if
   * both properties are present.
   *
   * By setting this value to the empty array `[]`, you are telling this
   * [[`TextWrap`]] to never match if any marks are found.
   *
   * @default null
   */
  validMarks?: string[] | null;
}

enum TextWrapCriteria {
  /**
   * Always add the closing character regardless of selection status.
   *
   * This will be default value
   */
  Always = 'always',

  /**
   * Only respond to the rule when the selection is empty.
   */
  EmptySelection = 'empty-selection',

  /**
   * Only respond to the rule when the selection is not empty
   */
  MultiSelection = 'multi-selection',

  /**
   * Never respond to the rule. Useful if you only want the rule to match
   * deletions.
   */
  Never = 'never',
}

enum TextWrapDeletions {
  /**
   * Auto remove both characters when there is no content between the character
   * and it's closing character.
   *
   * This is the default for backspace deletions.
   */
  ContentEmpty = 'content-empty',

  /**
   * Always auto delete the character and it's closing character when the are no
   * invalid characters between the two.
   */
  Always = 'always',

  /**
   * Never delete the matches.
   *
   * This is the default for forward deletions.
   */
  Never = 'never',
}

/** Return true to add the closing character */
type TextWrapCriteriaFunction = (selection: TextSelection, state: EditorState) => boolean;
```

- [ ] Support transformations like `[` => `[]`.
- [ ] Only support single characters. For multi character the advice will be to use input rules.
- [ ] Auto-remove on delete.
- [ ] Would be nice to also add some kind of animation for the deletions and insertions. This can be achieved with delayed commands that wrap the text which is to be added with a temporary class for 200ms before running the command to insert / delete the text. 
  - Thinking more deeply about this, I would love to also add this functionality to input rules. Instant changes can be jarring. Smooth transitions would be amazing to use.

### Context

For markup editors certain characters have meaning when wrapping content. In Markdown the `*` and `[` have such meaning. It's not trivial to add such rules to ProseMirror, but it would be great if it were possible to do so.

### LINKED ISSUE #714: Extract html string handling into a separate package
### Description

Right now the handling of html is built into core. I want to extract it into a separate extension. This extension will add a `getHtml` method to the context. It will also allow users to set the `stringHandler` option for their editor to be `html` rather than providing a specific function.

- [x] Rename `fromHtml` to `htmlToProsemirrorNode`.
- [x] Rename `toHtml` to `prosemirrorNodeToHtml`.
- [x] Create a new extension `@remirror/extension-html` which will be included by default in the core preset.

### LINKED ISSUE #715: New package `@remirror/extension-whitespace`
### Description

This plugin [`prosemirror-invisbles`](https://github.com/guardian/prosemirror-invisibles) hasn't been updated in a while. If it works, use it. If not, learn and integrate it into a new `@remirror/extension-whitespace` package.

### LINKED ISSUE #757: Add new package `@remirror/extension-meta`
### Description

In some use cases, it is desirable to implement custom behaviour off the back of commands raised by default Remirror extensions

For instance, if the `insertImage` command from the Image extension is used, it may be desired to upload the image to an API, and replace ` I'm thinking the best way to do this automatically is create an extension that basically wraps the `createCommands` method for every extension to intercepts the value returned, so when it returns true and `dispatch` is defined, it will do what you've done in the current pr and call `tr.setMeta` with the command being called, the extension name, and the args passed in. It should even be able to say if the command was called via a chained method, or without chaining.

> I'm not sure the performance penalty of this, so I'll likely create an external extension under `@remirror/extension-meta`

In regard to performance penalties, it may be desirable to add options that allow extensions to exclude, or include from these events, with inclusions taking precedence.

### LINKED ISSUE #778: `useSuggester` doesn't get updated on `setContent`
### Summary

An active suggestor doesn't get updated when the content is changed via `setContent`, leading to an inconsistent state. This happens independently if `triggerChange` is set for `setContent` or not.

### Steps to reproduce

1. Open reproduction case: https://codesandbox.io/s/hopeful-ganguly-wqbfo?file=/src/App.js
2. Enter a slash to trigger the suggestion
3. Click [Reset content] to change the content to something without a the trigger character (slash)
=> The suggestion is still active
4. Click again [Reset content]
=> Now the suggestion is no longer active

Here is a recording of the issue: https://discord.com/channels/726035064831344711/745695521305526302/776757457112006686

### Expected results

Suggestion is no longer active if the trigger character is no longer in place.

### Actual results

Suggestion stays active

### LINKED ISSUE #793: Add triple colon support `:::` for the CalloutExtension
### Description

- [x] Add support for typing `:::` to create a default `CalloutExtension`. 
- [x] Add support for typing `:::success` to create a named `CalloutExtension` with the class info.
- [ ] Add support for restricting the types of callouts that can be created.
- [ ] Add support for icons to the callout extension. 

### Context

The code would look very similar to the code used for the [`CodeBlockExtension`](https://github.com/remirror/remirror/blob/f7778ccb13efdfa5f6dffc39a50e347b2fc40a43/packages/%40remirror/extension-code-block/src/code-block-extension.ts#L284-L331
) for pressing the enter key after typing the `:::`. An input rule could also be used to handle pressing the space key as seen in the same code block extension. 

https://github.com/remirror/remirror/blob/f7778ccb13efdfa5f6dffc39a50e347b2fc40a43/packages/%40remirror/extension-code-block/src/code-block-extension.ts#L193-L201

### LINKED ISSUE #800: Escaping nodes doesn't work consistently for many node types
Currently if you enable the code extension and enter

```
`code` text
```

and then go to the beginning of the word `code`, press the left arrow and then type the text you enter is outside (before) the code block. This is desired; however this behaviour should also apply to bold, italic, links, etc.

Similarly codeblocks handle the left/up arrow to allow you to escape and add a plain paragraph above the code block; however this should also work for other block types: blockquotes, headings, bulleted list, images, etc.

### LINKED ISSUE #804: Add support for hover events to `@remirror/extension-events`
### Description

Right now hover events for nodes and marks aren't supported. 

Add support for mouseover events to capture the node / mark position as well as a `mouseleave` for the captured positions.

Add the hover event support to the `PositionerExtension` so that positions can be calculated based on hovering.

### LINKED ISSUE #820: Bad kerning on text selection due to <span class="selection">
### Summary

Remirror wraps selected text in `<span class="selection">` which breaks the text into multiple segments; this prevents the built in font kerning applying, which then causes later text to be moved to the right depending on what the kerning would have been at the beginning of the selection.

E.g. normally the `T` and `e` in `Te` will overlap slightly, but when we render `T<span class="selection">est</span>` the `Te` is forbidden from overlapping, so the `est` is moved to the right by a couple pixels.

### Steps to reproduce

Enter this into the playground; then slowly drag a selection from the end of the text backwards and watch the text "jitter."

```json
{
  "doc": {
    "type": "doc",
    "content": [
      {
        "type": "heading",
        "attrs": {
          "level": 1
        },
        "content": [
          {
            "type": "text",
            "text": "Test test test"
          }
        ]
      }
    ]
  },
  "playground": {
    "m": 0,
    "a": [],
    "e": [
      {
        "module": "remirror/extension/heading",
        "export": "HeadingExtension"
      }
    ],
    "p": []
  }
}
```

### Expected results

No jitter.

### Actual results

Jitter.

### Possible Solution

- remove `<span class="selection">` entirely
- disable kerning in Remirror (don't do this!)

### Screenshot(s)

jitter

jitter2

### LINKED ISSUE #823: LinkExtension – Characters interpreted as markdown when pasting link
### Summary

Pasting a link that contains valid markdown syntax causes unusually behavior, for example this link `https://www.reddit.com/r/aww/comments/k05w86/awoo_on_a_cold_day/` when pasted in the playground results in parts of the link becoming italic:

image

In my own usage of remirror it actually results in an infinite loop as `getMarkRange` [here](https://github.com/remirror/remirror/blob/next/packages/@remirror/extension-link/src/link-extension.ts#L433) returns an incorrect range:

image

(note that the end of the link is missing). So far I've been unable to reproduce this in the playground, but it seems like fixing the incorrect interpretation as markdown within the link would also fix the infinite loop even though it's not reproducible.

### Steps to reproduce

- Paste the link above into an editor with `LinkExtension` and `ItalicExtension`

### Expected results

- Link is not formatted

### Actual results

- Link contains italic text

### LINKED ISSUE #834: usePositioner doesn't support multiple editor instances
### Summary

Using the `usePositioner` hook when multiple editors are mounted on the page does not take into account the provider context. For [example](https://github.com/goabstract/block-editor-test/blob/use-positioner-bug/src/App.js#L29), floating menus positioned this way are positioned on the incorrect editor.

### Steps to reproduce

- Repo: https://github.com/goabstract/block-editor-test/tree/use-positioner-bug
- `git checkout use-positioner-bug`
- `yarn install`
- `yarn start`
- Select text in both editors

### Expected results

- One menu shows up at a time based on which editor is in focus, menu from one editor is not affected by state or positioned on another

### Actual results

- Menus from both editors are overlaid

### Screenshot(s)

Kapture 2020-12-21 at 14 24 59

### LINKED ISSUE #836: Add target="_blank" support in LinkExtension
### Description

There should be a way to enable `target="_blank"` for `LinkExtension`. Could be like this: 
```new LinkExtension({ target: '_blank' })```

### Context

My current solution is to create a custom extension that extends `LinkExtension`
```
// CustomLinkExtension.js
import { LinkExtension } from 'remirror/extension/link'
import { isElementDomNode, omitExtraAttributes } from 'remirror/core'

export class CustomLinkExtension extends LinkExtension {
  createMarkSpec(extra) {
    const AUTO_ATTRIBUTE = 'data-link-auto'
    return {
      attrs: {
        ...extra.defaults(),
        href: {},
        auto: { default: false }
      },
      inclusive: false,
      parseDOM: [
        {
          tag: 'a[href]',
          getAttrs: node => {
            if (!isElementDomNode(node)) {
              return false
            }

            const href = node.getAttribute('href')
            const auto =
              node.hasAttribute(AUTO_ATTRIBUTE) ||
              this.options.autoLinkRegex.test(node.textContent ?? '')
            return { ...extra.parse(node), href, auto }
          }
        }
      ],
      toDOM: node => {
        const { auto: _, ...rest } = omitExtraAttributes(node.attrs, extra)
        const auto = node.attrs.auto ? { [AUTO_ATTRIBUTE]: '' } : {}
        const rel = 'noopener noreferrer nofollow'
        const attrs = {
          ...extra.dom(node),
          ...rest,
          rel,
          target: '_blank',
          ...auto
        }

        return ['a', attrs, 0]
      }
    }
  }
}
```

### PR DESCRIPTION
**Pull request by ifiokjr**
### Description

Here's what's changed in the beta release.

- [x] Improved `react` API
- [x] Full `markdown` support with the `@remirror/extension-markdown` package.
- [x] Full formatting support
- [x] i18n support
- [x] A11y support for react via `reakit`
- [ ] Component Library (work in progress)
- [ ] Start adding experimental react native support (work in progress)
- [x] Todo list extension 

### Delayed

- ~Experimental svelte support~ - This will be added later in the year.
- ~New math extension~ (will be added after the beta is released)
- ~New pagination extension~ (will be added after the beta is released)
- ~New text wrap extension~ (will be added after the beta is released)

## Breaking

- Editor selection now defaults to the `end` of the document.
- Rename all `*Parameter` interfaces to `*Props`. With the exception of \[React\]FrameworkParameter which is now \[React\]FrameworkOptions.
- Remove `Presets` completely. In their place a function that returns a list of `Extension`s should be used. They were clunky, difficult to use and provided little to no value.
- Add core exports to `remirror` package
- Add all Extensions and Preset package exports to the `remirror/extensions` subdirectory. It doesn't include framework specific exports which are made available from `@remirror/react`
- Remove `remirror/react` which has been replaced by `@remirror/react`
- `@remirror/react` includes which includes all the react exports from all the react packages which can be used with remirror.
- Remove `@remirror/showcase` - examples have been provided on how to achieve the same effect.
- Remove `@remirror/react-social`
- Remove `@remirror/react-wysiwyg`
- Rename `useRemirror` -> `useRemirrorContext`
- Replace `useManager` with better `useRemirror` which provides a lot more functionality.
- Rename `preset-table` to `extension-tables`
- Rename `preset-list` to `extension-lists`. `ListPreset` is now `BulletListExtension` and `OrderListExtension`.
- New `createDecorations` extension method for adding decorations to the prosemirror view.
- Create new decorator pattern for adding `@commands`, `@helper` functions and `@keyBindings`.
- Deprecate `tags` property on extension and encourage the use of `createTags` which is a method instead.
- Add `onApplyState` and `onInitState` lifecycle methods.
- Add `onApplyTransaction` method.
- Rename interface `CreatePluginReturn` to `CreateExtensionPlugin`.
- Rewrite the `DropCursor` to support animations and interactions with media.
- Add support updating the doc attributes.
- Deprecate top level context methods `focus` and `blur`. They should now be consumed as commands
- Remove package `@remirror/extension-auto-link`.

### `ExtensionStore`

- Rename `addOrReplacePlugins` to `updatePlugins` in `ExtensionStore`.
- Remove `reconfigureStatePlugins` and auto apply it for all plugin updating methods.

One of the big changes is a hugely improved API for `@remirror/react`.

### `@remirror/extension-positioner`

- New `Rect` interface returned by the positioner `x: number; y: number; width: number; height: number;`
- Added `visible` property which shows if the position currently visible within the editor viewport.
- Improved scrolling when using the positioner.
- Fixed a lot of bugs in the positioner API.
- This DOMRect represents an absolute position within the document. It is up to your consuming component to consume the rect.
- `@remirror/react-components` exports `PositionerComponent` which internally
- Renamed the positioners in line with the new functionality.

```tsx
import React from 'react';
import { BoldExtension, CorePreset, ItalicExtension, MarkdownExtension } from 'remirror/extension';
import { Remirror, useRemirror } from '@remirror/react';

const Editor = () => {
  const { manager, onChange, state } = useRemirror({
    extensions: () => [new BoldExtension(), new ItalicExtension()],
    content: '<p><strong>I am strong.</strong> and <em>I am emphasized</em></p>',
    stringHandler: 'html',
  });

  return <Remirror manager={manager} onChange={onChange} state={state}></Remirror>;
};
```

The previous `useRemirror` is now called `useRemirrorContext` since it plucks the context from the outer `Remirror` Component. The `<RemirrorProvider />` has been renamed to `<Remirror />` and automatically renders an editor.

When no children are provided to the `<Remirror />` component it will automatically render a container `div` where the prosemirror editor will be placed. If you do add children it is up to you to import the `<EditorComponent />` and add it to the children or set the `autoRender` prop to `'start' | 'end' | true`.

`useManager` has been marked as `@internal` (although it is still exported) and going forward you should be using `useRemirror` as shown in the above example.

### `@remirror/extension-tables`

`@remirror/preset-table` is now `@remirror/extension-tables`. The `TableExtension` uses the new `createExtension` method to inject the `TableRowExtension` which in turn injects the `TableCellExtension` and `TableHeaderCellExtension`. To use tables in your editor the following is sufficient.

```tsx
import { TableExtension } from 'remirror/extensions';
import { Remirror, useRemirror } from '@remirror/react';

const Editor = () => {
  const { manager } = useRemirror(() => [TableExtension()]);

  return <Remirror manager={manager} />;
};
```

### UI Commands

- Add commands with UI configuration and **i18n** text labels
- `@command`, `@keyBinding`, `@helper` decorators for more typesafe configuration of extensions.
- `NamedShortcut` keybindings which can be set on the keymap extension. Currently `remirror` supports a google docs keyboard shortcuts.
- Add the `overrides` property to the extensions which for advanced users allows overriding of the default `NodeSpec` and `MarkSpec`.

### Caveats around inference

- Make sure all your commands in an extension are annotated with a return type of `CommandFunction`. Failure to do so will break all type inference wherever the extension is used.

  ```ts
  import { CommandFunction } from 'remirror';
  ```

- When setting the name of the extension make sure to use `as const` otherwise it will be a string and ruin autocompletion for extension names, nodes and marks.

  ```ts
  class MyExtension extends PlainExtension {
    get name() {
      return 'makeItConst' as const;
    }
  }
  ```

### `@remirror/react-hooks`

- Rename `useKeymap` to `useKeymaps`. The original `useKeymap` now has a different signature.

```tsx
import { useCallback } from 'react';
import { BoldExtension } from 'remirror/extensions';
import { Remirror, useHelpers, useKeymap, useRemirror, useRemirrorContext } from '@remirror/react';

const hooks = [
  () => {
    const active = useActive();
    const { insertText } = useCommands();
    const boldActive = active.bold();
    const handler = useCallback(() => {
      if (!boldActive) {
        return false;
      }

      return insertText.original('\n\nWoah there!')(props);
    }, [boldActive, insertText]);

    useKeymap('Shift-Enter', handler); // Add the handler to the keypress pattern.
  },
];

const Editor = () => {
  const { manager } = useRemirror({ extensions: () => [new BoldExtension()] });

  return <Remirror manager={manager} hooks={hooks} />;
};
```

- The `Remirror` component now has a convenient hooks props. The hooks prop takes an array of zero parameter hook functions which are rendered into the `RemirrorContext`. It's a shorthand to writing out your own components. You can see the pattern in use above.

### Commands

There are new hooks for working with commands.

- Each command has an `original` method attached for using the original command that was used to create the command. The original command has the same type signature as the `(...args: any[]) => CommandFunction`. So you would call it with the command arguments and then also provide the CommandProps. This is useful when composing commands together or using commands within keyBindings which need to return a boolean.

  - You can see the `insertText.original` being used in the `useKeymap` example above.

- `useCommands()` provides all the commands as hook. `useChainedCommands` provides all the chainable commands.

  ```tsx
  import { useCallback } from 'react';
  import { useChainedCommands, useKeymap } from '@remirror/react';

  function useLetItGo() {
    const chain = useChainedCommands();
    const handler = useCallback(() => {
      chain.selectText('all').insertText('Let it goo 🤫').run();
    }, [chain]);

    // Whenever the user types `a` they let it all go
    useKeymap('a', handler);
  }
  ```

### Dependencies

- Upgrade React to require minimum versions of ^16.14.0 || ^17. This is because of the codebase now using the [new jsx transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html).
- Upgrade TypeScript to a minimum of `4.1`. Several of the new features make use of the new types and it is a requirement to upgrade.
- General upgrades across all dependencies to using the latest versions.
  - All `prosemirror-*` packages.

### Issues addressed

- Fixes #569
- Fixes #407
- Fixes #533
- Fixes #652
- Fixes #480
- Fixes #566
- Fixes #453
- Fixes #508
- Fixes #715
- Fixes #531
- Fixes #535
- Fixes #536
- Fixes #537
- Fixes #538
- Fixes #541
- Fixes #542
- Fixes #709
- Fixes #532
- Fixes #836
- Fixes #834
- Fixes #823
- Fixes #820
- Fixes #695
- Fixes #793
- Fixes #800
- Fixes #453
- Fixes #778
- Fixes #757
- Fixes #804
- Fixes #504
- Fixes #714
- Fixes #37

### Checklist


Bertg:
May I suggest - for the public website - to go HAM?

Show an "all bells and whistles" (wysiwyg) instance of the editor on the front page, and impress the visitor with the full power of the editor. Today the initial impression is - let's be frank - a bit underwhelming. 

Even ProseMirror does it :)

ifiokjr:
This sounds like a great idea. I'll add something to the `beta` version of the website this weekend.

ocavue:
@whawker Could you try to post a comment in this pull request with the content `/` + `publish`. I want to test if you (as a member) can trigger a release in this pull request.

ocavue:
@whawker Thank you for doing that. It seems that it doesn't work. I will look into this.

PR #725, opened 2020-09-27
Fix: Don't extend annotation when typing at end of annotation

ronnyroeller:
### Description

Annotations auto-adjust as users enter content, e.g. an annotation grows if the user types in the middle of the annotation.
Incorrectly, the annotation also grew when the user added content directly after the annotation. Now, this leads to new, non-annotated content.

### Checklist

### DIFF
--- a/packages/@remirror/extension-annotation/src/annotation-plugin.ts
+++ b/packages/@remirror/extension-annotation/src/annotation-plugin.ts
@@ -39,7 +39,9 @@ export class AnnotationState<A extends Annotation = Annotation> {
       .map((annotation) => ({
         ...annotation,
         from: tr.mapping.map(annotation.from),
-        to: tr.mapping.map(annotation.to),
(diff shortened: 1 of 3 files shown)


packages/@remirror/extension-annotation/src/annotation-plugin.ts
```diff
@@ -39,7 +39,9 @@ export class AnnotationState<A extends Annotation = Annotation> {
       .map((annotation) => ({
         ...annotation,
         from: tr.mapping.map(annotation.from),
-        to: tr.mapping.map(annotation.to),
+        // -1 indicates that the annotation isn't extended when the user types
+        // at the end of the annotation
+        to: tr.mapping.map(annotation.to, -1),
```
ifiokjr, line 44: Good change, and the comment is helpful for future contributors. A+ 👍

ifiokjr (approved):

PR #739, opened 2020-10-01
Add `blockSeparator` option for multi-block annotations

ronnyroeller:
### Description

`getAnnotations()` returns all currently set annotations and enriches them with the text of the annotation. This can be used e.g. to show a list of all annotations outside the editor context.
Before, text from multiple blocks (in a multi-block annotation) was concatenated without any separated. Now, one can define via the `blockseparator` option which string to use as separator. For example, one could use a newline character to separate text from different blocks.

### Checklist

### DIFF
--- a/packages/@remirror/extension-annotation/src/types.ts
+++ b/packages/@remirror/extension-annotation/src/types.ts
@@ -1,3 +1,5 @@
+import type { AcceptUndefined } from '@remirror/core';
+
 export type GetStyle<A extends Annotation> = (
   annotations: Array<AnnotationWithoutText<A>>,
 ) => string | undefined;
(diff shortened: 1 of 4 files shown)


ifiokjr (approved):
Looks good

packages/@remirror/extension-annotation/src/types.ts
```diff
@@ -11,6 +11,16 @@ export interface AnnotationOptions<A extends Annotation = Annotation> {
...
+
+  /**
+   * Allows to format the text returned for each annotation.
+   *
+   * When `blockSeparator` is given, it will be inserted whenever a new
+   * block node is started.
+   *
+   * @see PromirrorNode.textBetween
+   */
+  blockSeparator?: string | undefined;
```
ifiokjr, line 23: To allow undefined as a value, the following needs to be added.

Add this to the top of the file.

```ts
import { AcceptUndefined } from 'remirror/core';
```

```suggestion
  blockSeparator?: AcceptUndefined<string>;
```

ifiokjr (approved):
Apologies for the delays. I'll update the `undefined` in a further commit.

Everything else looks good.

PR #761, opened 2020-10-20
Avoid `LinkExtension` `pasteRules` when `autoLink` enabled

whawker:
#### Issue

This fixes some inconsistent behaviour when using the `LinkExtension` with `autoLink: true`.

Pasting a link *without* `www.` prefix - `updateLink` command called, and `onUpdateLink` event fired
Pasting a link *with* `www.` prefix - `updateLink` command **not** called, and `onUpdateLink` event not fired

This is because the link created via the [paste rule](https://github.com/remirror/remirror/blob/next/packages/@remirror/extension-link/src/link-extension.ts#L245) was deemed to have [no changes](https://github.com/remirror/remirror/blob/next/packages/@remirror/extension-link/src/link-extension.ts#L438-L446) and would bail out before calling `updateLink`.

#### Change

This change prevents the paste rule from being applied, if `autoLink` is enabled, as auto linking will create the link regardless.

I have added this as breaking change, please let me know if that is incorrect.

### Checklist

### DIFF
--- a/.changeset/green-frogs-know.md
+++ b/.changeset/green-frogs-know.md
@@ -1,0 +1,5 @@
+---
+'@remirror/extension-link': major
+---
+
+**BREAKING** avoid using the paste rule if `autoLink` is enabled for `LinkExtension`
--- a/packages/@remirror/extension-link/src/__tests__/link-extension.spec.ts
+++ b/packages/@remirror/extension-link/src/__tests__/link-extension.spec.ts
@@ -448,7 +448,7 @@ describe('autolinking', () => {
     editorText += 'm';
     editor.insertText('m');
 
-    expect(onUpdateLink).toHaveBeenCalledWith('test.com', {
+    expect(onUpdateLink).toHaveBeenCalledWith(editorText, {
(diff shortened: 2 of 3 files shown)


whawker:
I tried to add a test for this, but I couldn't get the `paste` command of `jest-remirror` to trigger the paste rule.

ifiokjr:
Unfortunately, I haven't been able to figure out how to get `paste` doesn't work in `jsdom`. I spent some hours last week trying to fix it, but to no avail.

ifiokjr:
FYI, around this, I've created a new organization https://github.com/jest-runners which is my attempt at building a browser-based runner which relies on playwright for running tests in the supported browsers. 

`jsdom` is a great piece of technology, but I've lost days (maybe weeks), trying to figure out why a test that runs in the browser doesn't run in `jsdom` and visa versa. When working with keyboard input, element positioning and browser events, it falls short as an accurate representation of the intended usage environment.

ifiokjr (approved):
Looks good 👍

PR #767, opened 2020-11-05
Add support for Callouts

whawker:
### Description

This PR adds supports for callouts, coloured boxes that highlight particular content

*N.B.* I would like to add configurable icons (i.e.  ℹ️ / ⚠️) to these but unsure of the best approach to support SVG images, for example

### Checklist

### Screenshots

Screenshot 2020-11-05 at 16 37 25


packages/@remirror/extension-callout/src/callout-types.ts
```diff
@@ -0,0 +1,22 @@
...
+   * @default 'info'
+   */
+  defaultType?: string;
+}
+
+export interface CalloutAttributes extends ProsemirrorAttributes {
+  /**
+   * The type of callout, for instance info, warning, error or success. Defaults to info
+   */
+  type?: string;
```
whawker, line 21: Was unsure if this should be a `info/warning/error/success` enum, but figured a string might be more flexible for custom types like `critical`?

packages/@remirror/extension-callout/src/styles.ts
```diff
@@ -0,0 +1,26 @@
+import { css } from 'linaria';
+
+export const editorStyles = css`
+  div[data-callout-type] {
+    background: #eef6fc;
+    border-left: 2px solid #3298dc;
```
whawker, line 6: Default here is info styling, same as the default type.

whawker: Although on second thoughts, make this is unwise, as default type is configurable

packages/@remirror/extension-callout/src/callout-types.ts
ifiokjr: Yeah this makes sense 👍

packages/@remirror/extension-callout/src/styles.ts
ifiokjr: > Although on second thoughts, maybe this is unwise

I think agree with this sentiment. Since there will always be a default it's probably best to leave this part unstyled.

packages/@remirror/extension-callout/src/callout-types.ts
ifiokjr: If you want to have the best of both worlds, with support for autocompletion and also the ability to set any string you can use the following from `type-fest`.

```typescript
import { LiteralUnion } from 'type-fest';

export interface CalloutAttributes extends ProsemirrorAttributes {
  /**
   * The type of callout, for instance info, warning, error or success. Defaults to info.
   */
  type?: LiteralUnion<'info' | 'warning' | 'error' | 'success', string>;
}
```

packages/@remirror/extension-callout/src/callout-types.ts
```diff
@@ -0,0 +1,22 @@
...
+   * It is a property so it can change during the editor's life.
+   *
+   * @default 'info'
+   */
+  defaultType?: string;
+}
+
+export interface CalloutAttributes extends ProsemirrorAttributes {
+  /**
+   * The type of callout, for instance info, warning, error or success. Defaults to info
```
ifiokjr, line 19:
```suggestion
   * The type of callout, for instance `info`, `warning`, `error` or `success`.
   * 
   * @default 'info'
```

packages/remirror/package.json
```diff
@@ -48,6 +48,7 @@
     "@remirror/extension-bidi": "1.0.0-next.51",
     "@remirror/extension-blockquote": "1.0.0-next.51",
     "@remirror/extension-bold": "1.0.0-next.51",
+    "@remirror/extension-callout": "*",
```
ifiokjr, line 51:
```suggestion
    "@remirror/extension-callout": "0.0.0",
```

packages/@remirror/extension-callout/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "dependencies": {
+    "@babel/runtime": "^7.12.0",
+    "linaria": "^2.0.0"
+  },
+  "devDependencies": {
+    "@remirror/core": "1.0.0-next.51",
+    "@remirror/pm": "1.0.0-next.51"
+  },
+  "peerDependencies": {
+    "@remirror/core": "*",
```
ifiokjr, line 35: This is something I need to fix for the template extensions. But until I do, this will help prevent bugs.

```suggestion
    "@remirror/core": "1.0.0-next.51",
    "@remirror/pm": "1.0.0-next.51"
```

packages/@remirror/extension-callout/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "@remirror/pm": "*"
```
ifiokjr, line 36:
```suggestion
    "@remirror/pm": "1.0.0-next.51"
```

ifiokjr (changes requested):
This is brilliant! Thanks for putting in the time to create this 💚😄  

I've made a few minor suggestions. And I'll merge it as soon as they're addressed 👍

ifiokjr (approved):

packages/@remirror/extension-callout/src/callout-types.ts
```diff
@@ -0,0 +1,25 @@
+import { LiteralUnion } from 'type-fest';
```
ifiokjr, line 1:
```suggestion
import type { LiteralUnion } from 'type-fest';
```

ifiokjr (approved):

PR #780, opened 2020-11-14
Add new `@remirror/extension-codemirror` package

ocavue:
### Description

Fixes: #711 

### Checklist

- [x] I have read the [**contributing**](https://github.com/remirror/remirror/blob/HEAD/docs/contributing.md) document.
- [x] My code follows the code style of this project and `pnpm fix` completed successfully.
- [x] I have updated the documentation where necessary.
- [x] New code is unit tested and all current tests pass when running `pnpm test`.

### Screenshots

image

PR #781, opened 2020-11-14
Fix `updateMark` to return `false` when not applicable

whawker:
### Description

Fix `updateMark` so it returns `false` if the mark cannot be applied to the selection or range

### Checklist

### DIFF
--- a/.changeset/real-dancers-search.md
+++ b/.changeset/real-dancers-search.md
@@ -1,0 +1,5 @@
+---
+'@remirror/extension-link': patch
+---
+
+The behaviour of `commands.updateLink.isEnabled()` has been **fixed** to return `false` when the `link` mark can't be applied to the selection. This was fixed by a change in the `@remirror/core-utils` package.
--- a/packages/@remirror/core-utils/src/command-utils.ts
+++ b/packages/@remirror/core-utils/src/command-utils.ts
@@ -37,8 +37,26 @@ interface UpdateMarkParameter extends Partial<RangeParameter>, Partial<Attribute
 export function updateMark(parameter: UpdateMarkParameter): CommandFunction {
   return ({ dispatch, tr }) => {
     const { type, attrs = object(), appendText, range } = parameter;
-    const { selection } = tr;
-    const { from, to } = range ?? selection;
+    const selection = range ? TextSelection.create(tr.doc, range.from, range.to) : tr.selection;
+    const { $from, from, to } = selection;
+    let applicable = $from.depth === 0 ? tr.doc.type.allowsMarkType(type) : false;
+
+    tr.doc.nodesBetween(from, to, (node) => {
+      if (applicable) {
+        return false;
+      }
+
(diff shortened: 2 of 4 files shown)


packages/@remirror/core-utils/src/command-utils.ts
```diff
@@ -37,8 +37,26 @@ interface UpdateMarkParameter extends Partial<RangeParameter>, Partial<Attribute
 export function updateMark(parameter: UpdateMarkParameter): CommandFunction {
   return ({ dispatch, tr }) => {
     const { type, attrs = object(), appendText, range } = parameter;
-    const { selection } = tr;
-    const { from, to } = range ?? selection;
+    const selection = range ? TextSelection.create(tr.doc, range.from, range.to) : tr.selection;
+    const { $from, from, to } = selection;
+    let applicable = $from.depth === 0 ? tr.doc.type.allowsMarkType(type) : false;
```
whawker, line 42: Taken from [here](https://github.com/ProseMirror/prosemirror-commands/blob/3126d5c625953ba590c5d3a0db7f1009f46f1571/src/commands.js#L446-L457)

.changeset/real-dancers-search.md
```diff
@@ -0,0 +1,6 @@
+---
+'@remirror/core-utils': minor
+'@remirror/extension-link': minor
```
ifiokjr, line 3:
```suggestion
```

This line isn’t needed since the changed files relate to the tests only.

packages/@remirror/core-utils/src/command-utils.ts
ifiokjr: 👍

ifiokjr (approved):
Looks good 🙌  

One minor remark to update the changeset and once that’s in place I’ll merge.

.changeset/real-dancers-search.md
whawker: This PR does change behaviour of `commands.updateLink.isEnabled()` (only when the the link mark is not applicable) so could be relevant?

ifiokjr: That's a good point. 

Although, changes in `@remirror/extension-link` should already be captured by changesets in the individual package `CHANGELOG.md`.

However, I do see value in mentioning that the change provides a **known** bug fix to the link extension. I say, "known" because there are likely other places where the change has also had a positive impact. 

I'd prefer to capture that change in a separate `patch` changeset. This would mean that the changelog generated for the `@remirror/core-utils` package remains specific to what changed in the package.

The new changeset can be as follows:

```md
---
'@remirror/extension-link': patch
---

The behaviour of `commands.updateLink.isEnabled()` has been **fixed** to return `false` when the the `link` mark can't be applied to the selection. This was fixed by a change in the `@remirror/core-utils` package.
```

whawker: Done :+1:

ifiokjr:
I'll release this once #780 is ready. Also there's a few outstanding bugs which need to be addressed.

PR #807, opened 2020-11-21
fix(theme): remove `open-color` from dependencies

### LINKED ISSUE #806: Remove `open-color` from the dependencies
### Summary

I have recently updated my OS to macOS 10.15 (Catalina) and I found that I couldn't install `@remirror/theme`. The reason is:

1.  `@remirror/theme` depends on `open-color`
2. `open-color` depends on an old version of `npm`, which depends on an old version of `node-gyp`
3. `node-gyp` before v7 can't be installed on macOS 10.15 properly. `node-gyp` has a special [document](https://github.com/nodejs/node-gyp/blob/v7.1.2/macOS_Catalina.md) for this.

`open-color` has fixed this issue by [removing](https://github.com/yeun/open-color/commit/2b9342d19133eb391828799aa85d686093e1f5ec#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) `npm` from its dependencies. However, this patch was never been released in the whole year. Check https://github.com/yeun/open-color/issues/108 for more details.

Based on that, I think `open-color` is not a well-maintained project. So I propose that we remove `open-color` from the dependencies of `remirror`. What we import from `open-color` is just a JSON file and we can simply copy this file into `remirror` without installing `open-color`. `open-color` is under MIT, so this should be fine. 

If you agree with this idea, I can submit a PR to implement it.

### Steps to reproduce

N/A

### Expected results

N/A

### Actual results

N/A

### Possible Solution

Remove `open-color` from the dependencies.

### Screenshot(s)

N/A

ocavue:
### Description

Remove `open-color` from the dependencies.

Fixes #806 

### Checklist


ifiokjr (approved):
Looks good 👍

PR #825, opened 2020-12-09
fix: `getMarkRange` usage with mixed marks

tommoor:
### Description

If marks are mixed then `getMarkRange` does not return the entire mark length, instead stopping at the first change in mark (see included test). Fixing this should resolve the infinite loop I described seeing here https://github.com/remirror/remirror/issues/823

### Checklist

### DIFF
--- a/packages/@remirror/core-utils/src/__tests__/core-utils.spec.ts
+++ b/packages/@remirror/core-utils/src/__tests__/core-utils.spec.ts
@@ -8,6 +8,7 @@ import {
   p,
   pm,
   schema as testSchema,
+  strong,
   tableRow,
 } from 'jest-prosemirror';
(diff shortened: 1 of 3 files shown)


packages/@remirror/core-utils/src/__tests__/core-utils.spec.ts
```diff
@@ -237,6 +238,16 @@ describe('getMarkRange', () => {
     expect(range?.mark).toBeInstanceOf(Mark);
     expect(range?.text).toBe(expected);
   });
+
+  it('returns the entire range when marks are mixed', () => {
```
ifiokjr, line 242: Nice test :+1:

ifiokjr (approved):
Great stuff 😄👍

PR #830, opened 2020-12-17
fix: preserve attributes for `toggleBlockItem` and `setBlockType`

### LINKED ISSUE #819: toggleBlockItem, setBlockType wipe node attrs
### Summary

Many of the nodes in our implementation contain custom `attrs` such as `id`, for example. When using the inbuilt remirror commands such as `toggleHeading`, or it's sub command `toggleBlockItem` any custom attrs that were on the node are wiped away.

It's easy to see how this happens in the [implementation of setBlockType](https://github.com/remirror/remirror/blob/1adea88a600ea5f92f4403f6817a4acd140eb0b3/packages/%40remirror/core-utils/src/command-utils.ts#L188).

### Steps to reproduce

- use any of the `toggleX` commands such as `toggleHeading` in combination with custom attrs.

### Expected results

- `attrs` that aren't explicitly wiped are retained.

### Actual results

- `attrs` are removed from the node

### Possible Solution

I guess there's a chance that you might not consider this a bug, but it does mean that we're not able to use any of the inbuilt commands. Particularly with the `toggleX` commands I think it makes sense to retain any attrs that aren't explicitly passed an empty value.

Thoughts?

ifiokjr:
### Description

Add `getActiveNode` function which returns the information for an active node of the provided type.

Rename deprecated error constant from `COMMANDS_CALLED_IN_OUTER_SCOPE` to `SCHEMA`.

Add `getMarkRanges` which supports retrieving all the mark ranges from within the provided selection.

Preserve attributes when toggling a block node. This allows for better support when using `extraAttributes` and fixes [#819](https://github.com/remirror/remirror/issues/819).

Fixes #819 

### Checklist

### DIFF
--- a/packages/@remirror/core-utils/src/command-utils.ts
+++ b/packages/@remirror/core-utils/src/command-utils.ts
@@ -18,7 +18,7 @@ import { TextSelection } from '@remirror/pm/state';
 import { findWrapping, liftTarget } from '@remirror/pm/transform';
 
 import { getMarkRange, isMarkType, isNodeType } from './core-utils';
-import { isNodeActive, isSelectionEmpty } from './prosemirror-utils';
+import { getActiveNode, isNodeActive, isSelectionEmpty } from './prosemirror-utils';
 
 interface UpdateMarkParameter extends Partial<RangeParameter>, Partial<AttributesParameter> {
   /**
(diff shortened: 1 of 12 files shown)


packages/@remirror/core-utils/src/command-utils.ts
```diff
@@ -170,13 +173,20 @@ export function setBlockType(
 
       if (node.type === type) {
         applicable = true;
+        activeAttrs = node.attrs;
```
tommoor, line 176: Hey @ifiokjr – giving this functionality a try in v60 and it doesn't seem to be working as intended – my guess is this `if` statement, where it's checking that the node type === type before considering it's attributes, but by definition the `nodeType` is only the same in one toggle direction.

PR #848, opened 2021-01-20
fix: call setState in command in useEmoji

chaseadamsio:
I noticed while using the `useEmoji` hook with the `SocialEmojiComponent`  that when I click an emoji, the `state` persisted and the `SocialEmojiComponent` was left open even after selecting an emoji. As I started to look at the code, I noticed that `command` never sets the state to `null` for the `useEmojiChangeHandler`, and found that `setState` was called after the `return` in the `command`. 

This PR rearranges the `return` and call to `setState` so that the `state` is set to `null`.

### Notes

in the source code, `setState` was being called after `return` in the `useEmojiChangeHandler` `command` handler. 

That meant `setState` was unreachable and was stripped in the distributed files.

Not calling `setState` led to a situation where using the `SocialEmojiComponent` would remain open even after selecting an emoji.

From `use-emoji/dist/react-hooks.esm.js`:

```
  if (changeReason) {
      setState({
        list: emojiMatches,
        index: 0,
        command: (emoji, skinVariation) => {
          return _command(emoji, skinVariation); // no call to `setState`
        },
        range
      });
  }
```

### Checklist

### Screenshots

**Before**

Jan-19-2021 22-30-31

**After (adding `setState` to `react-hooks.browser.esm.js` bundled file)**

Jan-19-2021 22-32-40

### DIFF
--- a/packages/@remirror/react-hooks/src/use-emoji.ts
+++ b/packages/@remirror/react-hooks/src/use-emoji.ts
@@ -49,8 +49,8 @@ function useEmojiChangeHandler(setState: SetEmojiState) {
           list: emojiMatches,
           index: 0,
           command: (emoji, skinVariation) => {
-            return command(emoji, skinVariation);
             setState(null);
+            return command(emoji, skinVariation);
           },
           range,
         });


ifiokjr (approved):
Good spot! 👍

PR #849, opened 2021-01-27
feat(embed): set custom value for "type" of iframe embeds

ronnyroeller:
### Description

The "type" attribute for the iframe-extension indicates the kind of embed. Before, the attribute would allow only for values "youtube" and "custom" (youtube being supported natively by the iframe-extension).
This commit allows to set the "type" to any string. With this, one can distinglish between different kinds of custom embeds like "google-drive" vs. "vimeo".

This change is backwards-compatible.

Discussed here: https://discord.com/channels/726035064831344711/745695521305526302/80403347018298163

### Checklist

### DIFF
--- a/packages/@remirror/preset-embed/src/iframe-extension.ts
+++ b/packages/@remirror/preset-embed/src/iframe-extension.ts
@@ -1,5 +1,6 @@
 import { cx } from 'linaria';
 import { parse, stringify } from 'tiny-querystring';
+import type { LiteralUnion } from 'type-fest';
 
 import {
   ApplySchemaAttributes,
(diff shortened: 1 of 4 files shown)


packages/@remirror/preset-embed/src/iframe-extension.ts
```diff
@@ -36,7 +36,7 @@ export type IframeAttributes = ProsemirrorAttributes<{
   allowFullScreen?: 'true';
   width?: string | number;
   height?: string | number;
-  type?: 'custom' | 'youtube';
+  type?: 'youtube' | string;
```
ifiokjr, line 39: You can use `LiteralUnion` from `type-fest` which allows for autocompletion of the `youtube` string value.

```ts
type?: LiteralUnion<'youtube', string>;
```

ronnyroeller: Good one! Fixed via: https://github.com/remirror/remirror/pull/849/commits/f397e2861781dcc897a6919d3b19d21ac73acd2f

PR #860, opened 2021-02-12
chore(deps): add missing dependency cross-env

ocavue:
### Description

I found that in Github Codespace, the git hooks don't work well because of `cross-env-shell: not found`. 

```bash 
node ➜ ~/workspace/remirror (beta) $ git checkout -b fix_deps
Switched to a new branch 'fix_deps'
husky > post-checkout (node v14.15.4)
sh: 1: sh: 1: cross-env-shell: not foundcross-env-shell: not found

husky > post-checkout hook failed (cannot be bypassed with --no-verify due to Git specs)
```

*BTW, when I wrote this PR, I just found out that `cross-env` was in [maintenance mode](https://github.com/kentcdodds/cross-env/issues/257). `cross-env` will continue to support the latest Node.js version as we as to fix security/critical bugs, so it should be ok to continue to use it. I just post this as a small piece of news.*

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,6 +161,7 @@
     "case-anything": "^1.1.2",
     "core-js": "^3.8.2",
     "cpy-cli": "^3.1.1",
+    "cross-env": "^7.0.3",
     "delay": "^5.0.0",
     "detox": "^18.2.2",
     "downlevel-dts": "^0.7.0",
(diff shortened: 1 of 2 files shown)


ifiokjr (approved):
Good spot 👍

PR #862, opened 2021-02-13
chore: mutate peerDependencies when publishing PRs

### LINKED ISSUE #861: PR-based release for beta branch has wrong peerDependencies version
### Summary

https://unpkg.com/@remirror/core@0.0.0-pr706.2/package.json

image

### Steps to reproduce

N/A

### Expected results

N/A

### Actual results

N/A

### Possible Solution

N/A

### Screenshot(s)

N/A

ocavue:
### Description

Fixes #861

### DIFF
--- a/support/scripts/src/mutatate-packages.ts
+++ b/support/scripts/src/mutatate-packages.ts
@@ -48,6 +48,9 @@ export async function mutatePackageVersions(prerelease: string) {
     // Change the versions of internal devDependencies.
     mutateDependencies(versionedNames, json.devDependencies);
 
+    // Change the versions of internal peerDependencies.
+    mutateDependencies(versionedNames, json.peerDependencies);
+
     // log.debug(location, json.version);
     promises.push(limit(() => writeJSON(jsonLocation, json, options)));
   }
--- a/support/scripts/src/mutatate-packages.ts
+++ b/support/scripts/src/mutatate-packages.ts
@@ -61,6 +61,7 @@ export async function mutatePackageVersions(prerelease: string) {
   const rootJson = require(baseDir('package.json'));
   mutateDependencies(versionedNames, rootJson.dependencies);
   mutateDependencies(versionedNames, rootJson.devDependencies);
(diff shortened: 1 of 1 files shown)


ocavue:
Wait! This PR doesn't fix the issue:

https://unpkg.com/@remirror/core@0.0.0-pr862.1/package.json

image

ifiokjr (approved):

ocavue:
Not sure where did I make it wrong. The result is still incorrect.

https://unpkg.com/@remirror/core@0.0.0-pr862.2/package.json

Screenshot 2021-02-14 at 02 04 56


image

ifiokjr:
I think it's because it only runs the the action from the `default` branch. It doesn't use the PR version to prevent attacks from malicious PRs.

To test that your code works, I created a test file at `support/scripts/tmp.js`.

```js
const { mutatePackageVersions } = require('./');

mutatePackageVersions('pr862.3');
```

And then I ran `node support/scripts/tmp.js`.

This mutates all the package.json files in the repo and I can see that your changes fix the issue 👍 

Final point, I want to keep PR releases exclusive to changed code in packages which will be consumed by users. Not for testing internal build scripts, otherwise there's a lot of noise.

ocavue:
image

@ifiokjr I just realized that my behavior will change the target of `^0.0.0-pr706.2`. I should ask you before comment `/publish` in this PR. 😢

ifiokjr:
@ocavue that's not an issue actually. If someone is using the **PR** release or any of the pre-release versions all the dependencies should be pinned to an exact version.

I'm glad you released it to be honest. It's why you have contributor access to the repo. In the future if there's doubt, do what you feel is best with the following guidance.

- Releases are meant to be when changes have happened in public packages which can be consumed by users.
- An excessive number of releases in a short period of time, is not desirable. What is excessive? It's hard to say, but 5 releases in an hour might be a bit much. PR releases should be when there's something concrete that we want to make available to users for testing, after we already have some confidence that it works.

ocavue:
Thank you for the reminder. I will pay attention in the future.

Issue #847, opened 2021-01-19
Preview to-be-deleted cells in table


ronnyroeller:
### Description

The `table-extension` allows to delete rows, columns, or the table itself.

For better usability, the table should indicate the impact of a delete action before the user triggers it.

### Inspiration

Confluence highlights the to-be-deleted cells by changing their border color to red:
![image](https://user-images.githubusercontent.com/9339055/105056634-b851b200-5a74-11eb-8814-84fdfb4caa0b.png)

On hover, Confluence highlights also the delete icon around the border that would trigger the deletion:
![image](https://user-images.githubusercontent.com/9339055/105057548-9e649f00-5a75-11eb-9b2d-7f4a1393dac9.png)

Issue #846, opened 2021-01-19
Support cell background color


ronnyroeller:
### Description

Colors are an effective way to organize and visualize data in tables. For example, a status cell could use green for "Done" and red for "Behind schedule".

For this, the table should allow to assign a background to a cell (or multiple cells).

Ideally, this functionality is integrated in the cell context menu (https://github.com/remirror/remirror/issues/842).

### Inspiration

Confluence allow to set the background color via the cell popup menu:
![image](https://user-images.githubusercontent.com/9339055/105056198-39f51000-5a74-11eb-8322-9692e75c5db4.png)

Issue #844, opened 2021-01-19
One-click selection for all cells in a table row/column


ronnyroeller:
### Description

Users can select a full row in the table by selecting the first cell and the moving the mouse towards the last cell in the row.

To improve usability, users should be able to select a complete row/column by clicking on the border of the table.

### Inspiration

In Confluence, hovering over the border makes the section for a row/column clickable. Clicking it, selects all cells in the row/column:
![image](https://user-images.githubusercontent.com/9339055/105055044-ee8e3200-5a72-11eb-86ea-47e5dee2d733.png)
![image](https://user-images.githubusercontent.com/9339055/105055229-24cbb180-5a73-11eb-8a14-58b72a951d1c.png)

Issue #843, opened 2021-01-19
Allow to add & remove rows/columns from table border


ronnyroeller:
### Description

The `table-extension` allows to resize the column width with the mouse. In contrast, adding and removing rows/columns requires to call commands (e.g. from a menu).

For better usability, the table should allow to add/remove rows/columns from the table border.

### Inspriration

In Confluence, the [plus] button appears when hover over the dots around the table border - clicking it adds a row/column:
![image](https://user-images.githubusercontent.com/9339055/105053840-ad495280-5a71-11eb-958f-53066c442f47.png)
![image](https://user-images.githubusercontent.com/9339055/105053897-bc300500-5a71-11eb-8637-1503af98273f.png)

In Confluence, the [minus] button appears when the user selects a column/row - clicking it removes the row/column:
![image](https://user-images.githubusercontent.com/9339055/105054114-eaade000-5a71-11eb-8ec6-fb771fda4519.png)

PR #877, opened 2021-02-25
feat(extension-react-tables): new scoped package

ocavue:
### Description

This package includes two main features: table controllers and table cell menu. You can find more details about these feature in the [readme.md](https://github.com/remirror/remirror/tree/feature/react-tables/packages/remirror__extension-react-tables)

Fixes https://github.com/remirror/remirror/issues/842
Fixes https://github.com/remirror/remirror/issues/843
Fixes https://github.com/remirror/remirror/issues/844
Fixes https://github.com/remirror/remirror/issues/846
Fixes https://github.com/remirror/remirror/issues/847

### Checklist


ronnyroeller:
@ocavue One thing I realized: I'd expect editors to have only the simple or the advanced table extension - but not both.
This raises the question what's happening to data that contain the old table node. Is your plan to have this as a drop-in replacement (i.e. data created with the simple table extension can be read/edited with the new table extension)? Or are you more thinking of migrating the data when upgrading from simple to advanced table extension?

ocavue:
@ronnyroeller Sorry for the late response.

> Is your plan to have this as a drop-in replacement (i.e. data created with the simple table extension can be read/edited with the new table extension)?

Yes, that's my plan. I didn't delete anything from the original simple table extension so that the new extension should be able to both read and edit the original data. However, once the table is edited by the new extension, I expect the data may not work properly when going back to the old extension. This can be a problem when you have both old and new clients modifying the same data.

While this was my plan, I do need to write some tests to ensure that the serialization and deserialization would work as I expected.

ronnyroeller:
@ocavue That sounds excellent!

ocavue:
I can't understand this error when I run `pnpm build`:

```
$ pnpm build

🎁 info building bundles!
🎁 error @remirror/extension-react-tables "@remirror/extension-react-component" is imported by "../remirror__react-core/src/react-remirror.tsx" but the package is not specified in dependencies or peerDependencies
```

Some information:

1. `@remirror/extension-react-tables` does use `@remirror/react-core`
2. `@remirror/extension-react-tables` has include `@remirror/react-core` in its package.json's `dependencies`
3. `@remirror/extension-react-tables` doesn't directly use `@remirror/extension-react-component`

ocavue:
/publish

(this is just a test and no packages should be released)

ronnyroeller:
@ocavue I tried to integrate `pr877.3` into our app but ran into a couple of issues:

1. `@remirror/extension-react-tables` doesn't come with `@remirror/react#pr877.3`. I could solve this by installing extension-react-tables explicitely: `npm install -i @remirror/extension-react-tables`

2. Starting the app, I get many of those warnings:
   ```
   WARNING in ./node_modules/@remirror/extension-react-tables/dist/remirror-extension-react-tables.browser.esm.js 559:15-17
   "export 'cx' was not found in '@emotion/css'
   ...
   ```

3. When I add the `TableExtension` to my list of Remirror extensions, I get this error (in a webpack app):
   ```
   "_remirror_extension_react_tables__WEBPACK_IMPORTED_MODULE_23__.TableExtension is not a constructor"
   ```
   => This might be linked to the first point
   
Please let me know if you need more details.

ocavue:
@ronnyroeller Thanks for your response.

1. I will discuss with ifiokjr about it.
2. For current implementation, you need to install some emotion packages: 
3. I have no idea why this happened.

ocavue:
I will release a new version to fix the issue of `@rmotion/css` dependency after fixing a CI workflow bug.

ifiokjr:
@ocavue I'll work with you to remove the requirement for `@emotion/css`. It should be optional, and instead we should aim for a class-based approach. It's more work while developing, but hopefully a  better experience for end users.

ocavue:
Update: I create a super simple demo with the new package:

Source code: https://github.com/ocavue/remirror-vite-example
Live demo: https://remirror-extension-react-tables.vercel.app/

ronnyroeller:
For easier testing, I set up a CRA project with the table extension: https://github.com/ronnyroeller/cra-remirror-tables

It starts to look really smooth! 🎉 🎉 🎉 

A couple of findings will I tried it out:
- [ ] The popup menu doesn't close after selecting an option: [recoding](https://user-images.githubusercontent.com/9339055/111669310-f66e1680-8816-11eb-9c81-bb5d0a7ddaea.mov)
- [ ] Popup menu misses the option to remove the table
- [ ] The popup menu isn't available if multiple table cells are selected (I guess that's only important for changing background color).
- [ ] The text is on the cell bottom (would expect in the middle): 
image
- [ ] Clicking the upper-left corner selected the y-axis (I would have expected that this selects all cells): [recording](https://user-images.githubusercontent.com/9339055/111670399-16520a00-8818-11eb-9dcd-99ce5d3fe5b2.mov)

ocavue:
@ronnyroeller Thanks for your repo and findings. I'm happy to see that this package works fine with CRA/Webpack.

--- 

> The popup menu doesn't close after selecting an option

I will fix it in the next release.

--- 

> Popup menu misses the option to remove the table

I could add an option in the cell menu to remove the table. I just don't know if it's correct in UX. For example, Atlassian has two table menus: one for the cell and one for the whole table. They put the "remove button" feature into the table menu. Maybe we should follow their design?

WeChatWorkScreenshot_eee2fe3d-17fe-4b09-81d8-b157997b4d5e copy

--- 

> The popup menu isn't available if multiple table cells are selected (I guess that's only important for changing background color).

I think it's a good idea to make the popup menu available for multiple table cell selection. I do something want to delete multiple rows at once. I will fix it in the next release.

--- 

> The text is on the cell bottom (would expect in the middle):

I will fix the css issue in the next release.

--- 

> Clicking the upper-left corner selected the y-axis (I would have expected that this selects all cells)

This issue is tricky. I thought I fixed it but clearly I didn't fix all cases. I will dig into this and maybe I need to make some refactoring to totally fix it. 

---

I will update this comment once these issues are fixed.

ronnyroeller:
> I could add an option in the cell menu to remove the table

I think putting it in the popup is better as it's less opinionated (and I guess also easier to implement). Users could still create a separate delete panel if they wish to.

ocavue:
@ifiokjr 

I got a style issue when using storybook as a development environment. 

The class names in the storybook are seem to be some kind of `emotion` random class names. But the class names in the production build are well-formatted "remirror-xxxx". I'm not famous with `linaria` but I think this different expected. 

image


However, it seems like the class name `css-1lhrb7l` in storybook doesn't have the same style as the `remirror-editor` in the production build. For example the table cells in two environment have different hight.

ifiokjr:
@ocavue good spot

That's because the webpack config for storybooks swaps `linaria` with `emotion`. I'm happy to remove that if it works for you.

ronnyroeller:
@ocavue I asked our UX to test the PR. They reported that everything works nicely! 🎉 

The only thing she detected was that the background color of to-be-deleted cells doesn't change: https://github.com/remirror/remirror/issues/847

ocavue:
@ronnyroeller I think I missed this feature. I will add it ASAP.

examples/storybook-react/main.ts
```diff
@@ -31,11 +30,6 @@ export default {
     config.plugins = plugins;
     config.externals = externals;
 
-    if (process.env.NODE_ENV !== 'production') {
-      // Use emotion as an alias for linaria.
-      alias['@linaria/core'] = '@emotion/css';
-    }
-
```
ocavue, line 38: highlight for reviewers: linaria instead of emotion is used in storybook

packages/remirror__core-types/src/core-types.ts
```diff
@@ -208,21 +208,24 @@ export type DOMOutputSpec =
...
+      | 'attrs'
+      | 'selectable'
+      | 'draggable'
+      | 'code'
+      | 'defining'
+      | 'isolating'
+      | 'parseDOM'
+      | 'toDebugString'
+      | 'allowGapCursor'
+    >
```
ocavue, line 228: highlight for reviewers: I added `allowGapCursor` to `NodeExtensionSpec`

package.json
```diff
@@ -80,7 +81,8 @@
     "reset": "pnpm clean:all; rm pnpm-lock.yaml; pnpm symlink:root; pnpm install;",
     "run:prettier": "prettier --loglevel warn --ignore-path .eslintignore \"**/*.{js,jsx,ts,tsx,md,mdx,json,html,css,yml,yaml,graphql}\"",
     "size": "size-limit",
-    "storybook": "manypkg run examples/storybook-react start",
+    "storybook": "run-p fix:css:watch storybook:start",
```
ocavue, line 84: highlight for reviewers: Since we don't use emotion in storybook anymore, we have to run `pnpm fix:css` everytime when we change the theme. I made `pnpm storybook` to generate style files automagically when needed.

packages/remirror__theme/src/extension-tables-theme.ts
```diff
@@ -2,6 +2,15 @@ import { css } from '@linaria/core';
```
ocavue, line 2: highlight for reviewers: I added all the style for `@remirror/extension-react-tables` into `extension-tables-theme.ts`. Should I create a new file to handle this?

packages/remirror__theme/src/utils.ts
```diff
@@ -365,6 +365,7 @@ export const defaultRemirrorTheme: Remirror.Theme = {
     shadow3,
     backdrop: transparentize(foreground, 0.1),
     outline: transparentize(primary, 0.6),
+    divider: lighten(foreground, 0.8),
```
ocavue, line 368: highlight for reviewers: I added a new `divider` value into the `Remirror.Theme`

packages/remirror__theme/src/extension-tables-theme.ts
ifiokjr: Looks good

ronnyroeller:
I integrated material-ui to test the component customization (code: https://github.com/ronnyroeller/cra-remirror-tables). See screenshot below. All in all, this went super smooth! 🎉 

image

I stumbled over two points:
- `TableCellMenuButtonProps` and `TableCellMenuPopupProps` must be imported from `@remirror/extension-react-tables/dist/declarations/src/components/table-cell-menu`. It would be better to export them from top-level `@remirror/extension-react-tables`
- The popup overlaps the editor area when opened in the right most column. In below screenshot, this leads to a horizontal scrollbar to accommodate for the popop. In this case, it would be better to open the popup to the left as `<FloatingWrapper />` does (could we use something like `<FloatingWrapper />` here?):
Screen Shot 2021-04-01 at 10 19 05 PM

packages/remirror__extension-react-tables/src/views/table-view.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+import TableInsertButton, { shouldHideInsertButton } from '../components/table-insert-button';
+import { ReactTableNodeAttrs } from '../table-extensions';
+import { injectControllers } from '../utils/controller';
+import { setNodeAttrs } from '../utils/prosemirror';
+
+export type GetTableStyle = (attrs: ReactTableNodeAttrs) => { className: string; style: string };
+
+export function buildTableStyle(): GetTableStyle {
+  const previewSelectionBorderColor = '#0067ce';
```
ronnyroeller, line 17: Instead of hardcoding the color here: How about using one of the [existing theme colors](https://github.com/remirror/remirror/blob/beta/packages/remirror__styles/theme.css#L18) (or add a new color)?
This eases styling the tables in line with the rest of the editor.

packages/remirror__theme/src/extension-tables-theme.ts
```diff
@@ -2,6 +2,15 @@ import { css } from '@linaria/core';
 
 import { getTheme } from './utils';
 
+const controllerSize = 12;
+const previewSelectionBorderColor = '#0067ce';
```
ronnyroeller, line 6: Instead of hardcoding the colors, would it be better to define them as CSS variables in the theme (see https://github.com/remirror/remirror/blob/beta/packages/remirror__styles/theme.css#L18)? => e.g. `--preview-selection-border-color`.
This makes it easier to adjust those colors for users.

ronnyroeller:
@ocavue Would be good to add somewhere references which issues this PR is going to solve, so we keep track of what implemented them. Something like:
```
Fixes https://github.com/remirror/remirror/issues/842
Fixes https://github.com/remirror/remirror/issues/843
Fixes https://github.com/remirror/remirror/issues/844
Fixes https://github.com/remirror/remirror/issues/846
Fixes https://github.com/remirror/remirror/issues/847
```

ocavue:
@ronnyroeller Sorry for the late response. I was busy fixing your previous issue "Clicking the upper-left corner selected the y-axis (I would have expected that this selects all cells)" since I found the reason behind this issue. This issue has been fixed now.

---

> TableCellMenuButtonProps and TableCellMenuPopupProps must be imported from @remirror/extension-react-tables/dist/declarations/src/components/table-cell-menu. It would be better to export them from top-level @remirror/extension-react-tables

I exported these two types in `index.ts`.

--- 

> Instead of hardcoding the color here: How about using one of the existing theme colors (or add a new color)?
This eases styling the tables in line with the rest of the editor.

Good advice. I exported all colors with CSS variable now. Check the file `packages/remirror__styles/theme.css` in this PR for a quick overview of all the new CSS variables.

---

ifiokjr:
Really nice! Merging since the popup components will merged after this.

ocavue:
@ronnyroeller I just ask Ifiok to merge this PR since some other features are based on some modification in this branch. I will fix the issue of "Table cell menu overflow" with another PR.

ronnyroeller:
> I just ask Ifiok to merge this PR since some other features are based on some modification in this branch.

Awesome! 🎉 

> I will fix the issue of "Table cell menu overflow" with another PR.

I think this is an important one. I realize that it's pretty difficult to use the table right without it.

PR #882, opened 2021-03-05
feat(react-renderer): support static heading, iframe & link

ronnyroeller:
### Description

Added support for some common node types and marks.

The iframe and link handler can be parameterized to change the visualization vs the node attributes in RemirrorJSON. For example, it allows to open links in a new tab or set the iframe to a specific width/height (independently of how this was entered in the editor).

### Checklist

### DIFF
--- a/.changeset/twelve-berries-greet.md
+++ b/.changeset/twelve-berries-greet.md
@@ -1,0 +1,5 @@
+---
+'@remirror/react-renderer': minor
+---
+
+Support static rendering of heading, iframe, linkAdded support for some common node types and marks. The iframe and link handler can be parameterized to change the visualization vs the node attributes in RemirrorJSON. For example, it allows to open links in a new tab or set the iframe to a specific width/height (independently of how this was entered in the editor).
--- a/.changeset/twelve-berries-greet.md
+++ b/.changeset/twelve-berries-greet.md
@@ -2,4 +2,4 @@
 '@remirror/react-renderer': minor
 ---
 
-Support static rendering of heading, iframe, linkAdded support for some common node types and marks. The iframe and link handler can be parameterized to change the visualization vs the node attributes in RemirrorJSON. For example, it allows to open links in a new tab or set the iframe to a specific width/height (independently of how this was entered in the editor).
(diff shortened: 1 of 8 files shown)


packages/remirror__react-renderer/src/handlers/link.tsx
```diff
@@ -0,0 +1,18 @@
+import { FC } from 'react';
+import { Literal } from '@remirror/core';
+
+export const createLinkHandler = (overwriteProps?: Record<string, Literal>) => {
```
ronnyroeller, line 4: @whawker When you implemented the static renderer: Did you also have to overwrite attrs (e.g. to open links in new tab)? If so: Which pattern did you use to configure the node renderer?

whawker: Yes, I did have cause to overwrite in some cases, the pattern I used was similar to the handler map you have [here](https://github.com/remirror/remirror/blob/479fad906aebcf2168249ee65e0553543b6c9e19/packages/remirror__react-renderer/src/renderer.tsx#L24-L36)

If I required custom logic for a node type or mark, it renders a React component as a handler to support that, instead of rendering the default HTML node.

Does that answer the question, or have I missed something?

ronnyroeller: Thanks a lot. Very helpful!

.changeset/twelve-berries-greet.md
```diff
@@ -0,0 +1,5 @@
+---
+'@remirror/react-renderer': minor
+---
+
+Support static rendering of heading, iframe, linkAdded support for some common node types and marks. The iframe and link handler can be parameterized to change the visualization vs the node attributes in RemirrorJSON. For example, it allows to open links in a new tab or set the iframe to a specific width/height (independently of how this was entered in the editor).
```
ifiokjr, line 5:
```suggestion
Support static rendering of heading, iframe, link. Added support for some common node types and marks. The iframe and link handler can be parameterized to change the visualization vs the node attributes in RemirrorJSON. For example, it allows to open links in a new tab or set the iframe to a specific width/height (independently of how this was entered in the editor).
```

ifiokjr (approved):
Really good stuff here 💯

ronnyroeller:
🤦 I forgot to export the new handlers. I'll file another PR for that...

PR #885, opened 2021-03-08
ci: allow pr releases from a non-default branch

ocavue:
### Description

**This code is not tested.**

This PR allows remirror members to create a PR and release a pr-based version from this PR.

### Checklist

- [x] I have read the [**contributing**](https://github.com/remirror/remirror/blob/HEAD/docs/contributing.md) document.
- [x] My code follows the code style of this project and `pnpm fix` completed successfully.
- [x] I have updated the documentation where necessary.
- [x] New code is unit tested and all current tests pass when running `pnpm test`.

PR #889, opened 2021-03-12
ci: limit releases to owner, members and contributors

ocavue:
### Description

This PR includes three updates for the PR-based release workflow:

1. allow not only COLLABORATOR but also OWNER and MEMBER to release a PR-based version.
2. only allow users to submit exact "/publish" in the comment to avoid [this situation](https://github.com/remirror/remirror/pull/885#issuecomment-792465373).
3. Return an error when running this workflow on a forked repository.

---

In my view, one of the biggest challenges to write a CI workflow for an open-source project is preventing attacks. For this workflow as an example, an attacker may be able to override the `pnpm publish` command and upload the `NPM_TOKEN` to their own server.

We now have three layers to prevent this kind of attacks:

1. we limit the people who can trigger this workflow.
2. we don't allow people to run this workflow when the source branch is from a forked repository.
3. when using `actions/checkout`, we only provide `ref` but didn't provide `repository`, which means that we won't clone the code from a forked repository.

I think we should keep the existing protections for future iterations.

--- 

BYW, when I test this PR, I found that GitHub Action doesn't allow newline in the `if`. I just post it as a tip.

This is OK:

```yaml
jobs:
  publish_comment:
    runs-on: ubuntu-latest
    if: ${{ github.event.issue.pull_request && github.event.comment.body == '/publish' && (github.event.comment.author_association == 'OWNER' || github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'COLLABORATOR') }}
```

This is not OK:

```yaml
jobs:
  publish_comment:
    runs-on: ubuntu-latest
    if: |
      ${{ 
        github.event.issue.pull_request 
        && github.event.comment.body == '/publish' 
        && (
          github.event.comment.author_association == 'OWNER' 
          || github.event.comment.author_association == 'MEMBER' 
          || github.event.comment.author_association == 'COLLABORATOR'
        ) 
      }}
```

### Checklist

### DIFF
--- a/.github/workflows/publish-pr.yml
+++ b/.github/workflows/publish-pr.yml
@@ -9,7 +9,7 @@ env:
 jobs:
   publish_comment:
     runs-on: ubuntu-latest
-    if: github.event.issue.pull_request != '' && contains(github.event.comment.body, '/publish') && github.event.comment.author_association == 'COLLABORATOR'
+    if: ${{ github.event.issue.pull_request && github.event.comment.body == '/publish' && (github.event.comment.author_association == 'OWNER' || github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'COLLABORATOR') }}
     steps:
       - name: get pr information
         uses: actions/github-script@v3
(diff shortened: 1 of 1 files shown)


ifiokjr (approved):
Thanks for fixing my mistake 🤣

PR #898, opened 2021-03-19
docs: updated faq

eptdorado:
Updated FAQs. Added answers to following questions:
- Not able to run either beta nor next, tried some other too
- Why do I need an override parameter if I can already override the spec by using ES class extends syntax?
- Is using extends to override the behavior of an Extension (both for createNodeSpec and for other methods) a best practice for using remirror?
- Why does override only allow a subset of properties? For example, NodeSpecOverride doesn't include toDOM and fromDOM.
- Is react 17 supported?
- Is it possible to somehow retrieve the header level?
- Does anyone have a working example of embedding a youtube link => iframe? it seems like this is baked into preset-embed, but haven't gotten it to work yet.

### Checklist

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -26,3 +26,71 @@ import { htmlToProsemirrorNode } from 'remirror';
 
 const doc = htmlToProsemirrorNode({ html, schema: state.schema });
 ```
+
+### Not able to run either beta nor next, tried some other too
+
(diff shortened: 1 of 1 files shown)


ocavue:
Closing this PR since it's kind of outdated. But I still appreciate your contribution very much.

PR #905, opened 2021-04-01
feat(react-components): improve floating menu components

ifiokjr:
### Description

- Add new `MentionAtomPopupComponent`.
- Add `selectMark` command to the `CommandsExtension`.
- Add new events to `EventsExtension`. Also rename some exports.
- Fix `getSelectedWord` to only accept letters and numbers.
- Add `from` and `to` to the editor callback in `jest-remirror`.
- Prevent `LinkExtension` shortcut from automatically selecting the word.
- Refactor `FloatingMenu` component so that it can now support inputs within the editor via the `renderOutsideEditor` prop.
- Allow paste rules to replace the current selection.
- Add new `alwaysPositioner`.
- Support pasting a link to add a link to the selected text.

Make chainable commands lazy. Previously, chainable commands would immediately update the transaction. Now the transaction is only updated when `run` is called.

In this way the chainable command, now has no side effects until the `run` or `tr` methods are called.

Add `tr` method and `enabled` method to chainable commands.

- `enabled` checks to see whether the current chain can be run. Returns true if it is possible.
- `tr` runs all the transactions added without dispatching the command into the editor. Produces the side effects.


dalladirosa:
Thanks for the floating menu component it really works well. 
But there is a problem when I tried to use a minimum height for the editor, when the popout show it became like this

image
and when i try to scroll it looks like this 
image

What do you think?

ronnyroeller:
`pnpm storybook` on this PR fails currently with:
```
...
70% sealing React Docgen Typescript Plugin/remirror/node_modules/.pnpm/react-docgen-typescript@1.22.0/node_modules/react-docgen-typescript/lib/parser.js:475
            var trimmedText = (tag.text || '').trim();
                                               ^

TypeError: (tag.text || "").trim is not a function
    at /remirror/node_modules/.pnpm/react-docgen-typescript@1.22.0/node_modules/react-docgen-typescript/src/parser.ts:750:44
    at Array.forEach (<anonymous>)
    at Parser.getFullJsDocComment (/remirror/node_modules/.pnpm/react-docgen-typescript@1.22.0/node_modules/react-docgen-typescript/src/parser.ts:749:10)
```

ronnyroeller:
The `BulletListExtension` doesn't create new lines with bullets (build `0.0.0-pr905.2`). It does work with latest `0.0.0--pr706.17`.

Reproduction: https://codesandbox.io/s/remirror-pr-000-pr70617-forked-02p9g?file=/package.json

https://user-images.githubusercontent.com/9339055/114854439-1d534480-9de5-11eb-80d7-0ff95e5c2c97.mov

--- 

This might be fixed on beta (I'm not really sure the impact of @ocavue's commit based on the commit message): https://codesandbox.io/s/remirror-pr-000-pr70617-forked-02p9g?file=/package.json

PR #912, opened 2021-04-07
fix: update exports and `YjsExtension` style

ocavue:
### Description

This PR removes the `margin-top: 16px` style from the yjs extension, which will affect the style when using `all.css` event the users don't want to use the yjs extension.

This PR also exports some values.
 
### Checklist


ifiokjr (commented):
Thanks for this 👍

PR #890, opened 2021-03-13
feat(extension-checkbox): new package `@remirror/extension-checkbox`

whawker:
### Description

This is a draft pull request for work done so far, attempting to achieve #854 

I will comment below to high area that are incomplete and/or I could do with some help on

### Checklist

- [ ] I have read the [**contributing**](https://github.com/remirror/remirror/blob/HEAD/docs/contributing.md) document.
- [ ] My code follows the code style of this project and `pnpm fix` completed successfully.
- [ ] I have updated the documentation where necessary.
- [ ] New code is unit tested and all current tests pass when running `pnpm test`.

PR #920, opened 2021-04-14
feat(extension-list): implement list item with checkbox

ocavue:
### Description

This PR updates the origin `ListItemExtension` with two new attributes: `hasCheckbox` and `checked`. If a list item node has `hasChekcbox: true`, then it will render a checkbox at the beginning. This should be the simplest way to implement a togglable list since we don't need to add any new extensions. 

This PR also re-implements some commands from `prosemirror-schema-list` for better support to the checkbox.

A list item with a checkbox can be created by following input rules:

- Input `[ ] ` to create an unchecked list item.
- Input `- [ ] ` to create a normal list item firstly and then transform it to an unchecked list item. This matches the markdown style of writing.

In the above input rules, you can also use add `x` inside `[ ]` to directly create a checked list item.

The code is based on the origin implementation from #890. Big thanks to @whawker for letting me use some code from him and @ifiokjr for his customized commands help.

I would like to keep each PR smaller. Although I think this PR is ready for review and follow-up merging, we still have some things to follow up in the feature:

1. Markdown support
2. Menu support
3. When one list item is checked and the user press Enter, the next list item should not be checked too. I believe that it's a ProseMirror issue (https://github.com/ProseMirror/prosemirror/issues/1155).

### Checklist

### Screenshots

image

### DIFF
--- a/.changeset/strange-buses-play.md
+++ b/.changeset/strange-buses-play.md
@@ -1,0 +1,7 @@
+---
+'@remirror/extension-list': minor
+'@remirror/styles': minor
+'@remirror/theme': minor
+---
+
+Implement list item with checkbox.
--- a/.changeset/strong-emus-rule.md
+++ b/.changeset/strong-emus-rule.md
@@ -1,0 +1,5 @@
+---
+'@remirror/core-utils': patch
+---
+
+Use `tr.replaceRangeWith` instead of `tr.replaceWith` in inputRule.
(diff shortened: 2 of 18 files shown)


ocavue:
I'll fix the conflict soon

maccman:
Curious why this isn't a standalone checkbox component rather than attached to a list?

ocavue:
@maccman Just because it is simpler to implement. For example, it is easier to modify `listItemNode.attrs` than to insert a `checkboxNode` into a `listItemNode`. This can be helpful when modifying features such as `splitListItem` command.

On the other hand, do you think not having the checkbox as a standalone component will cause any problems? I have nothing in mind at the moment but I'd love to hear other people's opinions.

maccman:
@zbeyens what have you seen on slate-plugins around this? Does it make sense to have a standalone checkbox or is a checkbox-list the way to go?

zbeyens:
The slate-plugins list structure is 

```
ul
  li
    lic (custom)
    ul
      ...repeat
  ...repeat
```

where `lic` is the list item content which can be rendered as a checkbox or any custom component.

Issue #933, opened 2021-04-21
Pasting links ignores everything after "www."


ronnyroeller:
### Summary

Copy & pasting links in latest `pr706.19` is broken. It includes only the initial `www.`

### Steps to reproduce

1. Open www.google.com in browser tab & copy URL from browser bar 
2. Open in another tab the hosted demo for `p706`: https://remirror-gq9xpsij0-ocavue.vercel.app/?path=/story/link-extension--basic
3. Paste URL into demo

https://user-images.githubusercontent.com/9339055/115532768-64827f00-a296-11eb-98db-0c7e4ae06904.mov

### Expected results

URL is inserted

### Actual results

Only test `www.` is inserted

whawker:
I think this is related to the `(www\.)` match group in the [regex](https://github.com/remirror/remirror/blob/beta/packages/remirror__extension-link/src/link-extension.ts#L335) and the [`getMatchString`](https://github.com/remirror/remirror/blob/beta/packages/remirror__extension-link/src/link-extension.ts#L338) function.

As an aside, I think the regex should be the same as `this.options.autoLinkRegex` ?

ronnyroeller:
@whawker Thanks for pointer! Unfortunately, Github didn't send me your edits via email - so I already had a crack on the RegEx myself. On top of this, Github seems to have scrambled up your suggestion, so I can't really compare the solutions... :(

I totally agree that `autoLinkRegex` needs to match to avoid the `href` being out of sync with the link `text`,

I ended up with the following - basically making the existing groups non-capturing and adding a new capturing group covering the complete URL:
```diff
diff --git a/packages/remirror__extension-link/src/link-extension.ts b/packages/remirror__extension-link/src/link-extension.ts
index ec4309a2d..e6b2420de 100644
--- a/packages/remirror__extension-link/src/link-extension.ts
+++ b/packages/remirror__extension-link/src/link-extension.ts
@@ -163,7 +163,7 @@ export type LinkAttributes = ProsemirrorAttributes<{
     defaultProtocol: '',
     selectTextOnClick: false,
     openLinkOnClick: false,
-    autoLinkRegex: /(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[\da-z]+([.-][\da-z]+)*\.[a-z]{2,8}(:\d{1,5})?(\/\S*)?/,
[...]

PR #940, opened 2021-04-26
fix: pasting links ignores everything after "www."

ronnyroeller:
### Description

This happened because the RegEx used optional capturing groups.

The commit adds two improvements to capture the link text:
- All existing capturing RegEx groups are changed to be non-capturing. An additional capturing group covers now the complete URL.
- @whawker had pointed out that the RegEx to capture the link text wasn't linked to the RegEx to capture the link href (which can be changed by the user). This is solved by using the same, user-controlled, RegEx to capture the link href and text.

Fixes https://github.com/remirror/remirror/issues/933

### Checklist

### DIFF
--- a/packages/remirror__extension-link/src/link-extension.ts
+++ b/packages/remirror__extension-link/src/link-extension.ts
@@ -163,7 +163,7 @@ export type LinkAttributes = ProsemirrorAttributes<{
     defaultProtocol: '',
     selectTextOnClick: false,
     openLinkOnClick: false,
-    autoLinkRegex: /(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[\da-z]+([.-][\da-z]+)*\.[a-z]{2,8}(:\d{1,5})?(\/\S*)?/,
+    autoLinkRegex: /((?:http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[\da-z]+(?:[.-][\da-z]+)*\.[a-z]{2,8}(?::\d{1,5})?(\/\S*)?)/,
     defaultTarget: null,
   },
   staticKeys: ['autoLinkRegex'],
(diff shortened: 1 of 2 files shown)


packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -163,7 +163,7 @@ export type LinkAttributes = ProsemirrorAttributes<{
     defaultProtocol: '',
     selectTextOnClick: false,
     openLinkOnClick: false,
-    autoLinkRegex: /(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[\da-z]+([.-][\da-z]+)*\.[a-z]{2,8}(:\d{1,5})?(\/\S*)?/,
+    autoLinkRegex: /((?:http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[\da-z]+(?:[.-][\da-z]+)*\.[a-z]{2,8}(?::\d{1,5})?(\/\S*)?)/,
```
whawker, line 166: Does this work will full URL (including a path) as there is a capture group on `(\/\/S*)`

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -332,7 +332,7 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
     return [
       {
         type: 'mark',
-        regexp: /https?:\/\/(www\.)?[\w#%+.:=@~-]{2,256}\.[a-z]{2,8}\b([\w#%&+./:=?@~-]*)/gi,
+        regexp: this.options.autoLinkRegex,
```
whawker, line 335: The regex used here has the `g` and `i` flags

We should add them to the default `autoLinkRegex` so not to be breaking

packages/remirror__extension-link/src/link-extension.ts
whawker: I also think the beginning of this regex is a bit unnecessarily verbose.

I think we could just use

```suggestion
    autoLinkRegex: /(?:https?:\/\/)?[\da-z]+(?:[.-][\da-z]+)*\.[a-z]{2,8}(?::\d{1,5})?(?:\/\S*)?/gi,
```

ronnyroeller:
Thanks a lot for the review, @whawker !

PR #946, opened 2021-05-04
feat(extension-list): allow using command to create a checkbox list item

ocavue:
### Description

Make the command `toggleCheckboxChecked` not only change the `checked` attribute, but also transform a normal list item into a checkbox list item when necessary.

### Checklist

### DIFF
--- a/.changeset/serious-humans-hide.md
+++ b/.changeset/serious-humans-hide.md
@@ -1,0 +1,5 @@
+---
+'@remirror/extension-list': minor
+---
+
+Make the command `toggleCheckboxChecked` set the attribute `hasCheckbox`.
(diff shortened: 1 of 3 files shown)


ronnyroeller:
Look good to me

ronnyroeller:
@ocavue I though creating a new checkbox list should work something like:
```javascript
commands.toggleBulletList();
commands.toggleCheckboxChecked(false);
```

Doing so, `toggleCheckboxChecked` will jump out at `Make sure the list item is selected. Otherwise do nothing ` because `selection.node` is undefined.

Is above how this is supposed to be used?

ocavue:
@ronnyroeller let me try to re-implement checkbox as @whawker did and see if it's easy to export such a command.

PR #947, opened 2021-05-07
feat(extension-caption): work in progress - captions

whawker:
### Description

Adds the ability to add captions to compatible nodes

### Checklist

### Screenshots

https://user-images.githubusercontent.com/2003804/117468707-dd4b3180-af4c-11eb-961e-d5e8ebc07f89.mov

### DIFF
--- a/packages/remirror__extension-caption/src/figcaption-extension.ts
+++ b/packages/remirror__extension-caption/src/figcaption-extension.ts
@@ -1,0 +1,41 @@
+import {
+  ApplySchemaAttributes,
+  extension,
+  ExtensionTag,
+  NodeExtension,
+  NodeExtensionSpec,
+  NodeSpecOverride,
+} from '@remirror/core';
+
+export interface FigcaptionOptions {}
+
+/**
+ * An extension to annotate content with a caption
+ */
+@extension<FigcaptionOptions>({})
+export class FigcaptionExtension extends NodeExtension<FigcaptionOptions> {
+  get name() {
+    return 'figcaption' as const;
+  }
+
+  readonly tags = [ExtensionTag.Block];
+
+  createNodeSpec(extra: ApplySchemaAttributes, override: NodeSpecOverride): NodeExtensionSpec {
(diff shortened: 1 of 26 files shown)


packages/remirror__extension-image/src/image-extension.ts
```diff
@@ -94,6 +96,15 @@ export class ImageExtension extends NodeExtension<ImageOptions> {
...
+    // I don't like this at all, would very much appreciate feedback here
+    const hasCaption = this.store.extensions.find(
+      (e: AnyExtension) => e instanceof CaptionExtension,
+    );
+
+    if (hasCaption) {
+      return [ExtensionTag.BlockNode, ExtensionTag.Media, ExtensionTag.CaptionCompatible];
+    }
+
     return [ExtensionTag.InlineNode, ExtensionTag.Media];
```
whawker, line 108: @ifiokjr @ocavue would love some suggestions on how to improve this, I wonder if tags could be part of `nodeOverride` or something?

packages/remirror__extension-caption/src/figcaption-extension.ts
```diff
@@ -0,0 +1,41 @@
...
+export class FigcaptionExtension extends NodeExtension<FigcaptionOptions> {
+  get name() {
+    return 'figcaption' as const;
+  }
+
+  readonly tags = [ExtensionTag.Block];
+
+  createNodeSpec(extra: ApplySchemaAttributes, override: NodeSpecOverride): NodeExtensionSpec {
+    return {
+      content: 'paragraph',
```
whawker, line 25: Ideally this would just be a text node, but I can't get that to work. This seems to be the only way?

If it were an inline node, it would work well with the existing Image extension default, however it needs to be a block node to contain a paragraph (and hence text)

This also means that Image needs to become a block node, to avoid the "mixing block and inline nodes" error

devmrin:
Hi @whawker,

Is there any update on this?

PR #956, opened 2021-06-10
feat(extension-yjs): add ability to collaborate on annonations

whawker:
### Description

If both the YJS and annotations extensions are present in the editor, use the provided YDoc to share annotations via a YMap.

### Checklist

### DIFF
--- a/packages/remirror__extension-annotation/src/annotation-plugin.ts
+++ b/packages/remirror__extension-annotation/src/annotation-plugin.ts
@@ -1,4 +1,4 @@
-import { assertGet, TransactionProps } from '@remirror/core';
+import { assert, Transaction, TransactionProps } from '@remirror/core';
 import { Decoration, DecorationSet } from '@remirror/pm/view';
 
 import {
@@ -9,7 +9,7 @@ import {
   UpdateAnnotationAction,
 } from './annotation-actions';
 import { toSegments } from './annotation-segments';
-import type { Annotation, GetStyle, OmitText } from './annotation-types';
+import type { Annotation, GetStyle, MapLike, OmitText } from './annotation-types';
 
 interface ApplyProps extends TransactionProps {
   action: any;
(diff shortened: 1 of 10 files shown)


packages/remirror__extension-annotation/src/annotation-plugin.ts
```diff
@@ -24,7 +24,92 @@ export class AnnotationState<Type extends Annotation = Annotation> {
...
+  removeAnnotations(removeAction: RemoveAnnotationsAction): void {
+    removeAction.annotationIds.forEach((id) => {
+      this.map.delete(id);
+    });
+  }
+
+  setAnnotations(setAction: SetAnnotationsAction<Type>): void {
+    // YJS maps don't support clear
+    this.map.clear?.();
+    // eslint-disable-next-line prefer-arrow-callback
```
ankon, line 61: Out of interest: Why? You seem to be using arrow callbacks in other places?

whawker: Ah this is because the [third argument is `this`](https://github.com/yjs/yjs/blob/main/src/types/YMap.js#L175), so using an arrow function would bind the function's scope to the plugin's scope.

YJS's `forEach` does not provide a `thisArg` argument [like a native Map does](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach#syntax), so I can't

```ts
this.map.forEach((_, id, map) => {
  map.delete(id);
}, map);
```

Does that make sense? I suppose I could do this, but seems icky to me.

```ts
this.map.forEach((_, id, map) => {
  this.map.delete(id);
});
```

packages/remirror__extension-annotation/src/annotation-plugin.ts
```diff
@@ -24,7 +24,92 @@ export class AnnotationState<Type extends Annotation = Annotation> {
...
+
+  removeAnnotations(removeAction: RemoveAnnotationsAction): void {
+    removeAction.annotationIds.forEach((id) => {
+      this.map.delete(id);
+    });
+  }
+
+  setAnnotations(setAction: SetAnnotationsAction<Type>): void {
+    // YJS maps don't support clear
+    this.map.clear?.();
```
whawker, line 60: The `?.` should be redundant once https://github.com/yjs/yjs/pull/309 is released

packages/remirror__extension-annotation/src/annotation-plugin.ts
whawker: Ah, line 84... good spot. I'll look into it. Obviously I'll resolve the consistency as you know where I live 😨 😆

PR #960, opened 2021-06-15
feat(prosemirror-paste-rules): allow custom content for node

ocavue:
### Description

Update the API:

- For `type: "node"`: removed the `transformMatch` parameter and added a new optional `getContent` parameter, which is a function that transforms the match into content when creating the node. By using `getContent`, users can set content as not only a text node, but also `undefined` or something more complex.

### Checklist

### DIFF
--- a/.changeset/cold-parrots-sip.md
+++ b/.changeset/cold-parrots-sip.md
@@ -4,4 +4,4 @@
 
 Update the API:
 
-- For `type: "node"`: removed the `transformMatch` parameter and added a new `getContent` parameter, which can add custom content to the created node. By using `getContent`, users can set content as not only a text node, but also `undefined` or something more complex.
+- For `type: "node"`: removed the `transformMatch` parameter and added a new optional `getContent` parameter, which is a function that transforms the match into content when creating the node. By using `getContent`, users can set content as not only a text node, but also `undefined` or something more complex.
--- a/.changeset/cold-parrots-sip.md
+++ b/.changeset/cold-parrots-sip.md
@@ -1,5 +1,5 @@
 ---
-'prosemirror-paste-rules': minor
+'prosemirror-paste-rules': major
 ---
 
 Update the API:
--- a/.changeset/cold-parrots-sip.md
+++ b/.changeset/cold-parrots-sip.md
@@ -4,6 +4,4 @@
 
 Update the API:
 
-- for `type: "text"`: nothing change;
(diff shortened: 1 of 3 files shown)


packages/prosemirror-paste-rules/__tests__/paste-rules-plugin.spec.ts
```diff
@@ -1,5 +1,18 @@
...
+  hardBreak,
+  p,
+  schema,
+  strong,
+  table,
+  tableCell,
+  tableRow,
+} from 'jest-prosemirror';
 
+import { ProsemirrorNode } from '../../jest-prosemirror/node_modules/@remirror/pm/src';
```
ifiokjr, line 15: Update the import

```suggestion
import { ProsemirrorNode } from '@remirror/pm';
```

.changeset/cold-parrots-sip.md
```diff
@@ -0,0 +1,9 @@
+---
+'prosemirror-paste-rules': minor
+---
+
+Update the API:
```
ifiokjr, line 5: Add a brief explanation on why the change is necessary.

.changeset/cold-parrots-sip.md
```diff
@@ -0,0 +1,9 @@
+---
+'prosemirror-paste-rules': minor
```
ifiokjr, line 2: This is a breaking change. 

```suggestion
'prosemirror-paste-rules': major
```

ifiokjr (changes requested):
The code looks good 👍 

Can you just add an explanation for why the change is being made and what use-cases you have in mind?

packages/prosemirror-paste-rules/src/paste-rules-plugin.ts
```diff
@@ -279,6 +279,19 @@ export interface NodePasteRule extends BaseContentPasteRule {
...
    */
   nodeType: NodeType;
+
+  /**
+   * A function that transforms the match into the content to use when creating
+   * a node.
+   *
+   * Pass `() => {}` to remove the matched text.
+   *
+   * If this function is undefined, then the text node cutted from the match
```
ifiokjr, line 289:
```suggestion
   * If this function is undefined, then the text node is cut from the match
```

ifiokjr (approved):
Looks all good now 👍

PR #963, opened 2021-06-17
fix(core): fix onChange so `firstRender` only occurs once

whawker:
### Description

Context

```tsx
const handleChange = useCallback(({ firstRender }) => {
  if (firstRender) {
    doSomethingOnLoad();
  }
}, []);

return (
  <Remirror
    manager={manager}
    onChange={handleChange}
    autoRender
  />
);
```

Perhaps I'm mis-using the `firstRender` property of the `onChange` handler, but I expected `onChange` to be called with `firstRender` once?

I have found in practice this is not the case, due to the way private properties are transpiled, it actually creates a function `_classPrivateFieldSet(_this, _firstRender, false)` which I guess get pushed onto the stack and not immediately executed. This means onChange can be called multiple times with `firstRender = true`

This PR rearranges the code, so `onChange` with `firstRender = true` only happens once.

### Checklist

### DIFF
--- a/.changeset/swift-suns-push.md
+++ b/.changeset/swift-suns-push.md
@@ -1,0 +1,5 @@
+---
+'@remirror/core': patch
+---
+
+Ensure `onChange` callback is only called with `firstRender` once.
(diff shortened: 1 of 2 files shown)


.changeset/swift-suns-push.md
```diff
@@ -0,0 +1,5 @@
+---
+'@remirror/core': patch
+---
+
+Ensure onChange callback is only called with `firstRender` once
```
ifiokjr, line 5:
```suggestion
Ensure `onChange` callback is only called with `firstRender` once.
```

ifiokjr (approved):
Good spot 👍

PR #980, opened 2021-06-22
feat(extension-callout): add configurable emoji

Gaaaga:
### Description

Added configurable emoji to callouts. 

Added a new type: `blank`, so users are able to use a neutral background when they use the emoji as a semantic marker.

Fixes #871

### Checklist

- [x] I have read the [**contributing**](https://github.com/remirror/remirror/blob/HEAD/docs/contributing.md) document.
- [x] My code follows the code style of this project and `pnpm fix` completed successfully.
- [x] I have updated the documentation where necessary.
- [x] New code is unit tested and all current tests pass when running `pnpm test`.

### Screenshots

https://user-images.githubusercontent.com/16329365/123252988-7f37d500-d51f-11eb-9c7c-90eea5e607d2.mov

https://user-images.githubusercontent.com/24715727/123256602-bc05cb00-d523-11eb-8989-e11af894c3c4.mov

I truly appreciate @ocavue 's patient guidance, helping me dive right in the remirror project and have a better understanding of how it works.

PR #991, opened 2021-06-30
fix(callout-extension): improve the behavior of setting emoji

### LINKED ISSUE #871: Add configurable emoji to callouts
### Motivation

Today, callout nodes have a semantic marker: info, warning, error, success. Depending on the type, the callout can be styled. For example, an error callout could have a red background.

We found the 4 semantic categories too limiting for many use cases - especially "info" callouts could benefit from more, context-specific sub categories (see example below).

### Proposed solution

Users should be allowed to configure an emoji/icon for the callout freely, i.e. independent of the semantic marker.

### Example use cases

In these examples, the emoji/icon conveys better the message than a simple semantic callout:
image

Note: The emoji/icon allows users to create error/warning/success callouts (similar to semantic callouts):
image

### Inspirations

In Notion.so, clicking the emoji of the callout allows users to select a custom emoji:
image

Callout Ui design comes with icons for callouts: https://dribbble.com/shots/12882818-Callout-Ui-design-Figma-React-UI-kit

Gaaaga:
### Description

An improvement to the previous PR #980 , improving the behavior of setting emoji.

Fixes #871

### Checklist

### DIFF
--- a/packages/remirror__extension-callout/__stories__/callout.stories.tsx
+++ b/packages/remirror__extension-callout/__stories__/callout.stories.tsx
@@ -2,74 +2,138 @@ import 'remirror/styles/all.css';
 
 import { EmojiButton } from '@joeattardi/emoji-button';
 import { Blobmoji } from '@svgmoji/blob';
-import React, { useCallback } from 'react';
+import React, { useCallback, useEffect, useRef } from 'react';
 import { CalloutExtension } from 'remirror/extensions';
 import svgmojiData from 'svgmoji/emoji.json';
-import { EditorView, htmlToProsemirrorNode, ProsemirrorNode } from '@remirror/core';
+import { htmlToProsemirrorNode, ProsemirrorNode } from '@remirror/core';
 import { ProsemirrorDevTools } from '@remirror/dev';
-import { Remirror, ThemeProvider, useRemirror } from '@remirror/react';
+import {
+  Remirror,
+  ThemeProvider,
+  useCommands,
+  useRemirror,
+  useRemirrorContext,
+} from '@remirror/react';
 
 export default { title: 'Callouts' };
 
-const renderDialogEmoji = (node: ProsemirrorNode, view: EditorView, getPos: () => number) => {
-  const { emoji: prevEmoji, type } = node.attrs;
-  const emoji = document.createElement('span');
-  emoji.textContent = prevEmoji;
+const EmojiPicker = () => {
+  const pickerRef = useRef(new EmojiButton({ position: 'bottom', autoFocusSearch: false }));
+  const { updateCallout } = useCommands();
+  const { view } = useRemirrorContext();
+  const pos = useRef(-1);
 
-  const picker = new EmojiButton({
-    position: 'bottom',
-    autoFocusSearch: false,
-  });
+  const handleClickEmoji = useCallback(
+    (e: MouseEvent) => {
+      const target = e.target as HTMLElement;
+
+      e.preventDefault();
+
+      if (!target.matches('[data-emoji-container]')) {
+        return;
+      }
+
+      /**
+       * Find the document position of the click element.
+       */
+      pos.current = view.posAtDOM(target, 0);
+
+      pickerRef.current.togglePicker(target);
+    },
+    [view],
+  );
 
   /**
-   * Handle the selected emoji here
-   * Pass the new attributes to transaction to enable updating it from within the extension.
+   * Handle the selected emoji here.
+   * Use `updateCallout` commands to update new emoji.
+   * Need to pass pos information to commands, otherwise it will update the node where the cursor is located.
    */
-  picker.on('emoji', (selection) => {
-    const transaction = view.state.tr.setNodeMarkup(getPos(), undefined, {
-      emoji: selection.emoji,
-      type,
-    });
-    view.dispatch(transaction);
-  });
+  const handleSelectEmoji = useCallback(
+    (selection) => {
+      updateCallout({ emoji: selection.emoji }, pos.current);
+    },
+    [updateCallout],
+  );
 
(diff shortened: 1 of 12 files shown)


Gaaaga:
Hi @whawker ,can you help check this out. I found a command `updateCallout` in `CalloutExtension`, which is very close to what we are asking for, so I modified it to make it a bit more compatible, and so we can use it to modify `type` or `emoji` according to our requirements.

packages/remirror__extension-callout/__stories__/callout.stories.tsx
```diff
@@ -2,74 +2,138 @@ import 'remirror/styles/all.css';
...
+      }
+
+      /**
+       * Find the document position of the click element.
+       */
+      pos.current = view.posAtDOM(target, 0);
+
+      pickerRef.current.togglePicker(target);
+    },
+    [view],
```
whawker, line 43: Not a biggie, but I think you could use `[view.posAtDOM]` here?

packages/remirror__extension-callout/__stories__/callout.stories.tsx
```diff
@@ -2,74 +2,138 @@ import 'remirror/styles/all.css';
...
+const renderRandomEmoji = (node: ProsemirrorNode) => {
+  const { emoji: emojiCode } = node.attrs;
+  const emoji = document.createElement('img');
+  emoji.style.height = '48px';
+  emoji.style.width = '48px';
+  emoji.dataset.emojiContainer = '';
+  emoji.style.cursor = 'pointer';
+  emoji.alt = emojiCode;
+
+  const blobmoji = new Blobmoji({ data: svgmojiData, type: 'individual' });
```
whawker, line 135: Worth storing `new Blobmoji` in a ref?

packages/remirror__extension-callout/src/callout-utils.ts
```diff
@@ -24,12 +24,17 @@ export const dataAttributeEmoji = 'data-callout-emoji';
...
 ): attributes is CalloutAttributes {
-  return !!(
-    attributes &&
-    isObject(attributes) &&
-    isString(attributes.type) &&
-    attributes.type.length > 0
-  );
+  if (attributes && isObject(attributes)) {
+    if (isString(attributes.type) && attributes.type.length > 0) {
+      return true;
```
whawker, line 29: We `return true` before checking the emoji, so I think `isValidCalloutAttributes({ type: 'info', emoji: 97324 })` would return true, even thought the emoji is invalid?

Gaaaga: I think it allows us to just update the type value  without being wordily passing the emoji value(or just update emoji without mapping the existing type).

Gaaaga: Latter after this check, the existing value will be mapped in the attribute object if they are not provided. https://github.com/Gaaaga/remirror/blob/emojicallout-fix/packages/remirror__extension-callout/src/callout-utils.ts#L62

packages/remirror__extension-callout/__stories__/callout.stories.tsx
Gaaaga: The`exhaustive-deps` eslint rule prevents me from doing it, and if I assign `posAtDOM` to a new variable, it would cause a change to `this`, and making things a bit more tricky.
Do you have any idea to solve the problem?

packages/remirror__extension-callout/__stories__/callout.stories.tsx
Gaaaga: Because `renderRandomEmoji` is not a react functional component, it's a funtion called in createNoveView. So is it unable to use ref in it?

whawker: 🤦 sorry

packages/remirror__extension-callout/__stories__/callout.stories.tsx
whawker: I suppose this is only Storybook, so not a big deal - I just figured `view` is like to change a lot, although I guess as React compares with `Object.is` this is probably fine.

packages/remirror__extension-callout/src/callout-utils.ts
whawker: But I think the point still stands, `isValidCalloutAttributes({ type: 'info', emoji: 97324 })` will return `true` when surely it should return `false`?

Gaaaga: You are right! let me do some changes here.

packages/remirror__extension-callout/src/callout-utils.ts
```diff
@@ -24,12 +24,23 @@ export const dataAttributeEmoji = 'data-callout-emoji';
...
+    const attributesChecklist = Object.entries(attributes).map(([key, value]) => {
+      switch (key) {
+        case 'type':
+          return !!(isString(value) && value.length > 0);
+
+        case 'emoji':
+          return !!(isString(value) && value.length > 0);
+
+        default:
+          return true;
```
Gaaaga, line 37: @whawker Should I use return true in the default case? 
Which means the Callout extension can take attributes except 'type' and 'emoji' and will not constrain those value forms.

whawker: Yeah that make sense I think, with the `extraAttributes` functionality Remirror has.

whawker (approved):

whawker:
Nice work! Thanks for updating the functionality!

whawker:
Happy for me to merge @Gaaaga?

Gaaaga:
@whawker  Thanks! Feel free to merge this one :)

Issue #840, opened 2021-01-19
Resize embeds from remirror-iframe


ronnyroeller:
### Description

The `iframe-extension` renders an iframe to embed external content like YouTube videos.

The extension should allow users to resize the iframe. Depending on the embed, this could be with fixed aspect ratio (e.g. for YouTube videos) or with flexible aspect ratio (e.g. for Miro boards).

#### Fixed aspect ratio

Notion example for horizontal only resizing (see handlebars on left/right):
![image](https://user-images.githubusercontent.com/9339055/105040305-f7c2d300-5a61-11eb-8f05-d0f378e3f4f4.png)

#### Flexible aspect ratio

Notion example for horizontal and vertical resizing (see handlebars on left/right/bottom and in corners):
![image](https://user-images.githubusercontent.com/9339055/105040483-3062ac80-5a62-11eb-8f03-5124f8f10d51.png)

Gaaaga:
Currently working on this :)

ronnyroeller:
Fixed by @Gaaaga via https://github.com/remirror/remirror/pull/999

PR #999, opened 2021-07-11
feat(extension-iframe): make iframe resizable

Gaaaga:
### Description

- refactor of `ResizableNodeView`
- make iframe resizable by extend `ResizableNodeView`

See also #840

### Checklist

### Screenshots

https://user-images.githubusercontent.com/16329365/125412803-d1b73380-e3f1-11eb-8947-abb83be3be1e.mov


packages/remirror__extension-callout/src/callout-extension.ts
```diff
@@ -61,6 +61,7 @@ export class CalloutExtension extends NodeExtension<CalloutOptions> {
   createNodeViews(): NodeViewMethod {
     return (node: ProsemirrorNode, view: EditorView, getPos: boolean | (() => number)) => {
       const { type, emoji } = node.attrs;
+      console.log(node);
```
ocavue, line 64: Remove debug log

packages/prosemirror-resizable-view/package.json
```diff
@@ -37,7 +37,7 @@
   ],
   "dependencies": {
     "@babel/runtime": "^7.13.10",
-    "@remirror/core-helpers": "^1.0.0-next.60",
+    "@remirror/core": "1.0.0-next.60",
```
ocavue, line 40: Try not to do it. `prosemirror-resizable-view` is a package independent from the remirror. You can use this package in a non-remirror project. `@remirror/core-helper` is a simple dependency like `lodash` but `@remirror/core` is much heavier.

ocavue (changes requested):

packages/remirror__extension-embed/src/iframe-extension.ts
```diff
@@ -4,33 +4,23 @@ import {
...
-  defaultSource?: Static<string>;
-
-  /**
-   * The class to add to the iframe.
-   *
-   * @default 'remirror-iframe'
-   */
-  class?: Static<string>;
-}
+import { IframeOptions, ResizableIframeView } from './resizable-iframe-view';
```
ocavue, line 23: It may be a good idea to keep `IframeOptions` in `iframe-extension.ts` since (1) that's what most extensions do, and (2)  ·IframeOptions· plays more of a role as options for the Extension than as options for the nodeView.

packages/prosemirror-resizable-view/src/prosemirror-resizable-view.ts
```diff
@@ -1,172 +1,278 @@
...
+    /**
+     * Only initialize `handleBottom`, `handleBottomRight`, and `handleBottomLeft` in
+     * `ResizableRatioType.Flexible` mode
+     */
+    if (aspectRatio === ResizableRatioType.Flexible) {
+      const handleBottom = new ResizableHandle(ResizableHandleType.Bottom);
+      const handleBottomRight = new ResizableHandle(ResizableHandleType.BottomRight);
+      const handleBottomLeft = new ResizableHandle(ResizableHandleType.BottomLeft);
+
+      outer.addEventListener('mouseover', () => {
```
ocavue, line 90: Why do we only want to add `mouseover` listener when `aspectRatio === ResizableRatioType.Flexible`?

packages/prosemirror-resizable-view/src/prosemirror-resizable-view.ts
```diff
@@ -1,172 +1,278 @@
...
 
-      const onMouseMove = throttle(100, false, (e: MouseEvent) => {
-        const currentX = e.pageX;
+      handleBottomLeft.dom.addEventListener('mousedown', (e) =>
+        this.resizeHandler(e, view, getPos, handleBottomLeft),
+      );
+      outer.append(handleBottom.dom);
+      outer.append(handleBottomRight.dom);
+      outer.append(handleBottomLeft.dom);
+    }
```
ocavue, line 116: We could reduce the code by using some for loops:

```ts
    const types =
      aspectRatio === ResizableRatioType.Flexible
        ? [
            ResizableHandleType.Right,
            ResizableHandleType.Left,
            ResizableHandleType.Bottom,
            ResizableHandleType.BottomRight,
            ResizableHandleType.BottomLeft,
          ]
        : [ResizableHandleType.Right, ResizableHandleType.Left];

    const handles = types.map((type) => new ResizableHandle(type));

    for (const handle of handles) {
      handle.dom.addEventListener('mousedown', (e) => this.resizeHandler(e, view, getPos, handle));
      outer.append(handle.dom);
    }

    outer.addEventListener('mouseover', () => {
      handles.forEach((handle) => handle.setHandleDisplay(true));
    });

    outer.addEventListener('mouseover', () => {
      handles.forEach((handle) => handle.setHandleDisplay(true));
    });
```

packages/prosemirror-resizable-view/src/prosemirror-resizable-view.ts
```diff
@@ -1,172 +1,278 @@
...
+    outer.style.transition = 'width 0.15s ease-out, height 0.15s ease-out'; // make sure transition time is larger then mousemove event's throttle time
 
-        view.dispatch(transaction);
-      };
+    return outer;
+  }
 
-      document.addEventListener('mousemove', onMouseMove);
-      document.addEventListener('mouseup', onMouseUp);
+  resizeHandler(
```
ocavue, line 162: resizeHandler => resizeHandle

packages/prosemirror-resizable-view/src/prosemirror-resizable-view.ts
```diff
@@ -1,172 +1,278 @@
...
-      return Number.parseFloat(w);
-    }
+    return true;
   }
-
-  return 0;
 }
 
-function getSizeFromDom(element: HTMLElement | null): [number, number] {
+function getSizeFromDom(element: HTMLElement | null): DOMRect | null {
```
ocavue, line 268: We don't need `getSizeFromDom` anymore since we can just write one-line code: `element?. getBoundingClientRect ()`

packages/prosemirror-resizable-view/src/prosemirror-resizable-view.ts
```diff
@@ -1,172 +1,278 @@
...
+      this.resizeHandler(e, view, getPos, handleLeft),
+    );
+
+    outer.append(handleRight.dom);
+    outer.append(handleLeft.dom);
+
+    /**
+     * Only initialize `handleBottom`, `handleBottomRight`, and `handleBottomLeft` in
+     * `ResizableRatioType.Flexible` mode
+     */
```
ocavue, line 84: I'm thinking if we should give users more flexibility to choose handles. For example, what if a user only wants the two corner handles (bottomRight and bottomLeft)?

ocavue:
Great job! I left some review comments. I also want to ask @whawker to check this.

packages/prosemirror-resizable-view/src/prosemirror-resizable-view.ts
```diff
@@ -1,172 +1,278 @@
...
+      outer.style.width = node.attrs.width;
+      outer.style.height = node.attrs.height;
+    }
 
-        this.width = outer.style.width;
+    outer.style.maxWidth = '100%';
+    outer.style.minWidth = '50px';
+    outer.style.display = 'inline-block';
+    outer.style.lineHeight = '0'; // necessary so the bottom right handle is aligned nicely
+    outer.style.transition = 'width 0.15s ease-out, height 0.15s ease-out'; // make sure transition time is larger then mousemove event's throttle time
```
whawker, line 157: We seem to doing `x.style.y` an awful lot, Could we add a helper to core-utils that does something like

```ts
setStyle(outer, {
  maxWidth: '100%',
  minWidth: '50px',
  display: 'inline-block',
  lineHeight: '0',
  transition: 'width 0.15s ease-out, height 0.15s ease-out'
})
```

packages/prosemirror-resizable-view/src/resizable-view-handle.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  setHandleDisplay(visible?: true | string): void {
+    visible = visible || this.#handle.dataset.dragging;
+    this.#handle.style.opacity = visible ? '1' : '0';
+  }
+
+  dataSetDragging(data?: string): void {
+    this.#handle.dataset.dragging = data ?? '';
+  }
```
whawker, line 119: Expecting `'true'` as string seems like a implementation detail, I think this function should handle the data attribute string.

```suggestion
  dataSetDragging(data?: boolean): void {
    this.#handle.dataset.dragging = data ? 'true' : '';
  }
```

packages/remirror__extension-embed/__stories__/iframe.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <AddIframe />
+      </Remirror>
+    </ThemeProvider>
+  );
+};
+
+const AddIframe = () => {
+  const commands = useCommands();
+  const [src, setSrc] = useState('https://remirror.io/');
+  const onclick = () => commands.addIframe({ src, height: 250, width: 500 });
```
whawker, line 29:
```suggestion
  const handleClick = () => commands.addIframe({ src, height: 250, width: 500 });
```

whawker (commented):
Looks good to me! Just a few minor code style suggestions, but otherwise looks good to me!

packages/prosemirror-resizable-view/src/prosemirror-resizable-view.ts
```diff
@@ -1,172 +1,278 @@
...
+  /**
+   * `createElement` - a method to produce the element DOM element for this prosemirror node.
+   * The subclasses have to implement this abstract method.
+   */
+  abstract createElement(props: {
+    node: ProsemirrorNode;
+    view: EditorView;
+    getPos: () => number;
+    options?: OptionShape;
+  }): HTMLElement;
```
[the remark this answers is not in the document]
ocavue: @whawker I would like to get some advice from a senior developer. We have two API designs here and I don't know how to choose. Which one is better in your mind?

```ts
class ResizableNodeView {
  constructor(xx, xx, xx, createElement) {
    this.createElement = createElement
  }
}
```

vs 


```ts
abstract class ResizableNodeView {
  constructor(xx, xx, xx) {
  }

  abstract createElement()
}
```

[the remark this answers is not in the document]
ocavue: @whawker I would like to get some advice from a senior developer. We have two API designs here and I don't know how to choose. Which one is better in your mind?

```ts
class ResizableNodeView {
  constructor(xx, xx, xx, createElement) {
    this.createElement = createElement
  }
}
```

vs 


```ts
abstract class ResizableNodeView {
  constructor(xx, xx, xx) {
  }

  abstract createElement()
}
```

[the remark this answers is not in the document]
whawker: I'm my opinion, the second is much more extendable.

The first adds more and more arguments to the constructor, only some of which may be required.

The second could add further abstract methods, - as an aside we could avoid `abstract` altogether? Perhaps we define a default implementation that can be overridden when extended.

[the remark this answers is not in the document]
whawker: A third option is something like the `Positioner.fromPositioner` pattern? But that might not be relevant here?

P.s. @ifiokjr said he was planning on reimplementing node views?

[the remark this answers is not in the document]
ocavue: There is not breaking change for node view so we can implement the feature with the old way for now.

packages/prosemirror-resizable-view/src/prosemirror-resizable-view.ts
Gaaaga: Good tips! I think we could use `Object.assign()` here.

packages/prosemirror-resizable-view/src/prosemirror-resizable-view.ts
Gaaaga: Because in different ResizableRatioType the horizontal handles have different behavior.  So I think we could better just offer several general options, and the arguments could be neat. 
And we could extend any general type in the future, if needed.

packages/prosemirror-resizable-view/src/prosemirror-resizable-view.ts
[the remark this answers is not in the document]
Gaaaga: @whawker  I can't think of any child node to be implemented here as default node ,that make sense.
Because I think it is designed as a wrapper container,  that has to be extended with thoese node views who want to be resizable.

Gaaaga:
@ocavue @whawker Sorry for the late reply and thanks for your reviews!
I had made some chages and left a few comments :)

packages/create-context-state/__tests__/tsconfig.json.3424582592
```diff
@@ -0,0 +1,37 @@
+{
```
whawker, line 1: Just curious, are all these tsconfig files normal? Specifically the `tsconfig.json.<number>` file suffix, the number on the end looks odd to me.

packages/prosemirror-resizable-view/src/prosemirror-resizable-view.ts
```diff
@@ -1,172 +1,238 @@
...
-      handle.dataset.mouseover = 'true';
-      setHandleDisplay(true);
+      handles.forEach((handle) => handle.setHandleVisibility(true));
     });
 
     outer.addEventListener('mouseout', () => {
-      handle.dataset.mouseover = '';
-      setHandleDisplay();
+      handles.forEach((handle) => handle.setHandleVisibility(false));
     });
```
whawker, line 86: Should we add a [`destroy`](https://prosemirror.net/docs/ref/#view.NodeView.destroy) handler, to remove these event listeners?

packages/prosemirror-resizable-view/src/prosemirror-resizable-view.ts
whawker: Looks like we added `Object.assign` everywhere, except here? 

P.s. I still think we should add a dedicated `setStyle` helper (that internally uses `Object.assign`), rather than using `Object.assign` directly.

packages/remirror__extension-embed/__stories__/iframe.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <AddResizableIframe />
+      </Remirror>
+    </ThemeProvider>
+  );
+};
+
+const AddResizableIframe = () => {
+  const commands = useCommands();
+  const [src, setSrc] = useState('https://remirror.io/');
+  const onclick = () => commands.addIframe({ src, height: 250, width: 500 });
```
whawker, line 60:
```suggestion
  const handleClick = () => commands.addIframe({ src, height: 250, width: 500 });
```

packages/remirror__extension-image/src/resizable-image-view.ts
```diff
@@ -1,21 +1,24 @@
...
+
+  createElement({ node }: { node: ProsemirrorNode }): HTMLImageElement {
+    const inner = document.createElement('img');
+    inner.setAttribute('src', node.attrs.src);
+    inner.style.width = '100%';
+
+    inner.style.minWidth = '50px';
+    inner.style.minHeight = '30px';
+
+    inner.style.objectFit = 'contain'; // maintain image's aspect ratio
```
whawker, line 21: Could use the `setStyle` helper here?

packages/create-context-state/__tests__/tsconfig.json.3424582592
Gaaaga: I use `pnpm fix` command to correct typecheck, and it generated these files.  
Should I do something else to fix them?

packages/remirror__core-utils/src/dom-utils.ts
```diff
@@ -13,6 +13,16 @@ export function getStyle(
   return style.getPropertyValue(property);
 }
 
+/**
+ * Set more styles to the given element.
+ */
+export function setStyle(
+  target: HTMLElement,
+  styles: Partial<CSSStyleDeclaration>,
+): Partial<CSSStyleDeclaration> {
```
whawker, line 22: 🙌

whawker (approved):
Approved, thank you for this. Great work @Gaaaga!

ocavue (approved):

PR #1005, opened 2021-07-17
chore: remove github codespaces configuration

ocavue:
### Description

Remove the Docker image for GitHub Codespaces. By using the default environment instead of a Docker container, we can get more pre-installed tools. GitHub Codespaces users will need to install `pnpm` manually but it should be easy.

### Checklist


ifiokjr:
Can't we update the devcontainer?

ifiokjr (approved):
Okay, let's do it. It should be trivial to re-add if needed.

ocavue:
Sorry for the late response. I sleep too much these days.

Compare to the default GitHub Codespaces container, the old container we used (`mcr.microsoft.com/vscode/devcontainers/javascript-node`) lacks some tools, e.g. `docker`. Let's say someone want to use Docker to start a WebRTC server and test the YJS feature, then it's better to use the default container.

PR #1020, opened 2021-07-24
ci: keep the lockfile in the changeset workflow

ocavue:
### Description

This could cause some workflow issues if we recreate the lockfile

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,11 +99,10 @@
     "preversion:changeset": "node support/scripts/changeset-forced-update",
     "version:changeset": "changeset version",
     "postversion:changeset": "node support/scripts/changeset-forced-update --clean",
-    "version:ci": "run-s version:changeset version:date version:repo fix:prettier version:lockfile",
+    "version:ci": "run-s version:changeset version:date version:repo fix:prettier",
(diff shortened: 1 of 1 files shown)


ifiokjr:
@ocavue what issues does this fix?

ocavue:
@ifiokjr I was trying to fix this CI workflow [error](https://github.com/remirror/remirror/runs/3150329506#step:6:10) in #1019

In `packages/remirror__xxxx/__stories__/tsconfig.json`, there will be something like this:

```
"@storybook/react": [
"../../../node_modules/.pnpm/@storybook+react@6.3.4_dfad392d5450b8683a621f3ec486af19/node_modules/@storybook/react/types-6-0.d.ts"
],
```

This path could be changed if the `pnpm-lock.yaml` get update, which will cause `pnpm typecheck` failed to pass. I don't know why but it seems that `pnpm fix` won't fix this without `pnpm prune`.

Beside this reason, re-generate `pnpm-lock.yaml` could cause some potential errors because not all dependencies follow the `semver`. Only use `package.json`s can make sure the code runs as expected.

PR #1031, opened 2021-08-01
feat(extension-file): first version

ocavue:
### Description

TODO:

- [ ] add more document
- [x] make the default react component implement better looking
- [x] add an example to show how to upload files via browsers' file dialog
- [ ] make `IframeExtension` and `ImageExtension` has preview while uploading in the background. 

### Checklist

### Screenshots

https://user-images.githubusercontent.com/24715727/127881972-1142c65d-ce65-48eb-87ea-4aa783e93647.mp4

### DIFF
--- a/packages/remirror__extension-file/__stories__/file.stories.tsx
+++ b/packages/remirror__extension-file/__stories__/file.stories.tsx
@@ -1,0 +1,153 @@
+import 'remirror/styles/extension-file.css';
+
+import React, { useCallback, useEffect } from 'react';
+import { ProsemirrorDevTools } from '@remirror/dev';
+import { Remirror, ThemeProvider, useCommands, useRemirror } from '@remirror/react';
+
+import { FileExtension } from '../src';
+import { createBaseuploadFileUploader } from '../src/file-uploaders/bashupload-file-uploader';
+import { createObjectUrlFileUploader } from '../src/file-uploaders/object-url-file-uploader';
+import { createSlowFileUploader } from '../src/file-uploaders/slow-file-uploader';
+
+export default { title: 'File extension' };
+
+export const Default = (): JSX.Element => {
+  const extensions = useCallback(() => [new FileExtension({})], []);
+  const { manager, state } = useRemirror({ extensions, content, stringHandler: 'html' });
+
+  return (
+    <>
+      <p>
+        Default Implementation. Uses <code>FileReader.readAsDataURL</code> under the hood.
+      </p>
+      <ThemeProvider>
+        <Remirror manager={manager} initialContent={state} autoRender>
+          <UploadFileButton />
+          <ProsemirrorDevTools />
+        </Remirror>
+      </ThemeProvider>
+    </>
+  );
+};
+
+Default.args = {
+  autoLink: true,
+  openLinkOnClick: true,
+};
+
+export const WithObjectUrl = (): JSX.Element => {
+  const extensions = useCallback(
+    () => [new FileExtension({ uploadFileHandler: createObjectUrlFileUploader })],
+    [],
+  );
+  const { manager, state } = useRemirror({ extensions, content, stringHandler: 'html' });
+
+  return (
+    <>
+      <p>
+        Uses <code>URL.createObjectUrl</code> under the hood.
+      </p>
+      <ThemeProvider>
+        <Remirror manager={manager} initialContent={state} autoRender>
+          <UploadFileButton />
+          <ProsemirrorDevTools />
+        </Remirror>
+      </ThemeProvider>
+    </>
+  );
+};
+
+export const WithBashupload = (): JSX.Element => {
+  const extensions = useCallback(
+    () => [new FileExtension({ uploadFileHandler: createBaseuploadFileUploader })],
+    [],
+  );
+  const { manager, state } = useRemirror({ extensions, content, stringHandler: 'html' });
+
+  return (
+    <>
+      <p>
+        Actually upload the file to <a href='https://bashupload.com/'>https://bashupload.com/</a>.{' '}
+        bashupload is an open-source project created by IO-Technologies. Star it on{' '}
+        <a href='https://github.com/IO-Technologies/bashupload'>GitHub</a>.
+      </p>
+      <ThemeProvider>
+        <Remirror manager={manager} initialContent={state} autoRender>
+          <UploadFileButton />
+          <ProsemirrorDevTools />
+        </Remirror>
+      </ThemeProvider>
+    </>
+  );
+};
+
+export const UploadProgress = (): JSX.Element => {
+  const extensions = useCallback(
+    () => [new FileExtension({ uploadFileHandler: createSlowFileUploader })],
+    [],
+  );
+  const { manager, state } = useRemirror({ extensions, content, stringHandler: 'html' });
+
+  return (
+    <>
+      <p>
+        An example with slow uploading speed. You can see the upload progress and an abort button in
+        this example.
+      </p>
+      <ThemeProvider>
+        <Remirror manager={manager} initialContent={state} autoRender>
+          <UploadFileButton />
+          <ProsemirrorDevTools />
+        </Remirror>
+      </ThemeProvider>
+    </>
+  );
+};
+
+function useFileDialog() {
+  const [files, setFiles] = React.useState<FileList | null>(null);
+
+  const openFileDialog = () => {
+    const input = document.createElement('input');
+    input.type = 'file';
+    input.multiple = true;
+
+    input.addEventListener('change', (event: Event) => {
+      const { files } = event.target as HTMLInputElement;
+      setFiles(files);
+    });
+
+    input.click();
+  };
+
+  return { files, openFileDialog };
+}
+
+function useInsertFile() {
+  const { files, openFileDialog } = useFileDialog();
+  const { uploadFiles } = useCommands();
+
+  useEffect(() => {
+    if (files) {
+      const fileArray: File[] = [];
+
+      for (const file of files) {
+        fileArray.push(file);
+      }
+
+      uploadFiles(fileArray);
+    }
+  }, [files, uploadFiles]);
+
+  return { openFileDialog };
+}
+
+const UploadFileButton: React.FC = () => {
+  const { openFileDialog } = useInsertFile();
+  return <button onClick={openFileDialog}>Upload file</button>;
+};
+
+const html = String.raw; // Just for better editor support
+const content = html`<p>
+  Drag and drop one or multiple non-image files into the editor or click the button at the bottom.
+</p>`;
--- a/packages/remirror__extension-file/src/file-component.tsx
+++ b/packages/remirror__extension-file/src/file-component.tsx
@@ -1,0 +1,155 @@
+import React, { useEffect, useState } from 'react';
+import { NodeViewComponentProps, useCommands } from '@remirror/react';
+import {
+  CloseLineIcon as StopIcon,
+  DeleteBin6LineIcon as DeleteIcon,
+  DownloadLineIcon as DownloadIcon,
+  File2LineIcon as FileIcon,
+} from '@remirror/react-components/all-icons';
+import { ExtensionFileTheme } from '@remirror/theme';
+
+import type { FileAttributes } from './file-extension';
+import { UploadContext } from './file-upload-context';
+
+export type FileComponentProps = NodeViewComponentProps & {
+  context?: UploadContext;
+  abort: () => void;
+};
+
+export const FileComponent: React.FC<FileComponentProps> = ({
+  node,
+  getPosition,
+  context,
+  abort,
+}) => {
+  const attrs = node.attrs as FileAttributes;
+  const fileSize = formatFileSize(attrs.fileSize ?? 0);
+  return (
+    <div className={ExtensionFileTheme.FILE_ROOT}>
+      <IconButton>
+        <FileIcon />
+      </IconButton>
+      <div className={ExtensionFileTheme.FILE_NAME}>{attrs.fileName}</div>
+      <div className={ExtensionFileTheme.FILE_SIZE}>{fileSize}</div>
+      <div style={{ flex: 1, minWidth: '8px' }} />
+
+      {context ? (
+        <UploadingFileAction context={context} abort={abort} />
+      ) : (
+        <UploadedFileAction attrs={attrs} getPosition={getPosition as () => number} />
+      )}
+    </div>
+  );
+};
+
+const UploadingFileAction: React.FC<{ context: UploadContext; abort: () => void }> = ({
+  context,
+  abort,
+}) => {
+  const progress = useProgress(context);
+
+  return (
+    <>
+      <div className={ExtensionFileTheme.FILE_UPLOAD_PROGRESS}>{progress}</div>
+      <StopButton abort={abort} />
+    </>
+  );
+};
+
+const UploadedFileAction: React.FC<{
+  attrs: FileAttributes;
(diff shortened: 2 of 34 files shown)


packages/remirror__extension-file/src/file-extension.tsx
```diff
@@ -0,0 +1,355 @@
...
+        type: 'file',
+        fileHandler: (props) => {
+          let pos: number | undefined;
+
+          if (props.type === 'drop') {
+            pos = props.pos;
+          }
+
+          for (const file of props.files) {
+            this.startLoadFile(file, pos);
```
whawker, line 130: I think we should add an option to this extension that allows you to filter out certain file types. For instance you may wise to ignore images file drops, as they will be handled by the image extension.

packages/remirror__extension-file/src/file-extension.tsx
```diff
@@ -0,0 +1,355 @@
...
+  }
+
+  @command()
+  deleteFile(pos: number): CommandFunction {
+    return ({ tr, state, dispatch }) => {
+      const node = state.doc.nodeAt(pos);
+
+      if (node && node.type === this.type) {
+        if (dispatch) {
+          dispatch(tr.delete(pos, pos + 1).scrollIntoView());
```
whawker, line 177: Think we should provide an `onFileDeleted` or maybe even `tr.setMeta`, so a user can delete files from remote storage when the node is removed from the document.

whawker:
```suggestion
          dispatch?.(tr.delete(pos, pos + 1).scrollIntoView());
```

packages/remirror__extension-file/src/file-extension.tsx
```diff
@@ -0,0 +1,355 @@
...
+  }
+
+  @command()
+  renameFile(pos: number, fileName: string): CommandFunction {
+    return ({ tr, state, dispatch }) => {
+      const node = state.doc.nodeAt(pos);
+
+      if (node && node.type === this.type) {
+        if (dispatch) {
+          dispatch(tr.setNodeMarkup(pos, undefined, { ...node.attrs, fileName }));
```
whawker, line 194:
```suggestion
          dispatch?.(tr.setNodeMarkup(pos, undefined, { ...node.attrs, fileName }));
```

whawker (commented):
Nice work! Apologies I haven't had chance to run this locally yet, but added a few initial thoughts.

packages/remirror__extension-file/src/file-extension.tsx
ocavue: > or maybe even `tr.setMeta`

I don't understand this part. Could you give more details about this?

whawker: I was thinking if we set meta on the transaction, then a user could write a plugin to detect those transactions and perform additional behaviour - i.e. deleting the actual file from their S3 bucket etc.

```ts
tr.setMeta(this.name, { action: 'DELETE', url: node.attrs.url })
dispatch?.(tr.delete(pos, pos + 1).scrollIntoView());
```

We could even expose an `onDeleteFile` handler option for this use case, [similar to how `onUpdateLink` works](https://github.com/remirror/remirror/blob/beta/packages/remirror__extension-link/src/link-extension.ts#L644-L662).

packages/remirror__extension-file/src/file-extension.tsx
ocavue: added a new option `pasteRuleRegexp`

packages/remirror__extension-file/src/file-extension.tsx
ocavue: I added a new `onDeleteFile` handler. Hope this option can cover most use cases. If users have some other requirement, then maybe they could override or wrap the `deleteFile` command.

packages/remirror__extension-file/src/file-extension.tsx
```diff
@@ -0,0 +1,375 @@
...
+  findPlaceholderPos,
+  placeholderPlugin,
+  setPlaceholderAction,
+} from './file-placeholder-plugin';
+
+export interface FileOptions {
+  loadFileHandler?: LoadFileHandler;
+  getAttrsBeforeUpload?: GetAttrsBeforeUpload;
+  getAttrsAfterUpload?: GetAttrsAfterUpload;
+  render?: (props: FileComponentProps) => React.ReactElement<HTMLElement> | null;
```
Gaaaga, line 39: Could render return a `HTMLElement` but not `React.ReactElement`?

ocavue: Users can use a React Component or a ProseMirror NodeView, because both of them have life cycle functions. But HTMLElement doesn't have life cycle functions. If users want to use NodeView, then just extend the FileExtension and implement the `createNodeView` method and remove `ReactComponent` should be simple enough.

Life cycle functions are needed because the component needs to register a callback function to the `FileLoader` so that the component can be updated when the upload progress is updated.

packages/remirror__extension-file/src/file-placeholder-plugin.ts
```diff
@@ -0,0 +1,71 @@
...
+    pos: number;
+  };
+  remove?: {
+    id: unknown;
+  };
+}
+
+interface PlaceholderPluginData {
+  set: DecorationSet;
+  playloads: Map<unknown, any>;
```
whawker, line 17: Possibly typo?

```suggestion
  payloads: Map<unknown, any>;
```

packages/remirror__extension-file/src/file-loader.ts
```diff
@@ -0,0 +1,45 @@
...
+   * used to initiate the read operation.
+   */
+  readonly result: string | ArrayBuffer | null;
+
+  /**
+   * Aborts the read operation.
+   */
+  abort: () => void;
+
+  addEventListener: (
```
whawker, line 27: Loving the adapter pattern here, however I think `addEventListener`, `loadend` etc is too closely coupled to `FileReader`'s implementation, perhaps we should be a bit more generic.

For instance [CKEditor has a nice adapter pattern](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/upload-adapter.html#the-anatomy-of-the-adapter) where the only things you need to implement are `upload` and `abort`. 

I'm not saying this is exactly the interface we should implement, just an example of a more generic design.

packages/remirror__extension-file/src/file-extension.tsx
```diff
@@ -0,0 +1,375 @@
...
+      if (node && node.type === this.type) {
+        dispatch?.(tr.setNodeMarkup(pos, undefined, { ...node.attrs, fileName }));
+        return true;
+      }
+
+      return false;
+    };
+  }
+
+  private preLoadFile(id: FileId, file: File, pos: number | undefined): Promise<FileLoader> | void {
```
whawker, line 222: I find these method names slightly confusing, I think we're looking for something like

`preLoadFile` -> `createFilePlaceholder`

`postLoadFile` -> `onFileLoaded`

`startLoadFile` -> `uploadFile`

packages/remirror__extension-file/src/file-extension.tsx
```diff
@@ -0,0 +1,375 @@
...
+
+    this.store.view.dispatch(tr);
+
+    // return a promise that resolves when the file is loaded, whether
+    // successfully or unsuccessfully
+    return new Promise<FileLoader>((resolve) => {
+      fileLoader.addEventListener('loadend', () => {
+        resolve(fileLoader);
+      });
+    });
```
whawker, line 249: Not sure this belongs here, think we should just return the `fileLoader` here, and do this in `startLoadFile` instead?

packages/remirror__extension-file/src/file-loaders/bashupload-file-loader.ts
```diff
@@ -0,0 +1,110 @@
...
+        fileName: file.name,
+        fileType: file.type,
+        fileSize: file.size,
+      };
+    },
+    getAttrsAfterUpload: (_: File, fileLoader: FileLoader) => {
+      const url = (fileLoader as SimpleFormFileLoader).result;
+      const error = (fileLoader as SimpleFormFileLoader).error;
+      return { url: isString(url) ? url : '', error: error ? error.message : null };
+    },
```
whawker, line 108: Perhaps these methods should be part of the FileLoader adapter interface?

packages/remirror__extension-file/src/file-extension.tsx
```diff
@@ -0,0 +1,375 @@
...
+      ...this.options.getAttrsAfterUpload(file, fileLoader),
+      id: null,
+    };
+    tr.setNodeMarkup(placeholderPos, undefined, attrs);
+    view.dispatch(tr);
+  }
+
+  private startLoadFile(file: File, pos?: number | undefined): void {
+    // A fresh object to act as the ID for this upload.
+    const id = {};
```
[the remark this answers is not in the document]
ocavue: I stole this line of code and its comment from the official example: https://prosemirror.net/examples/upload/

When we use an empty object as an ID, we actually use its memory address. This brings two advantages over using a unique string as an id:

1. Eliminate the possibility of duplicate IDs (Although the probability is very small)
2. Increase performance since we don't need a complex algorithm to generate a string

ocavue:
@whawker I implemented a new `FileUploader` interface. Check [file-uploader.ts](https://github.com/remirror/remirror/blob/file-extension/packages/remirror__extension-file/src/file-uploader.ts) for the interface and check [default-file-uploader.ts](https://github.com/remirror/remirror/blob/file-extension/packages/remirror__extension-file/src/file-uploaders/default-file-uploader.ts) for one implementation.

packages/remirror__extension-file/src/file-loader.ts
ocavue: Check the [comment](https://github.com/remirror/remirror/pull/1031#issuecomment-894934050) for a new API

packages/remirror__extension-file/__stories__/file.stories.tsx
```diff
@@ -0,0 +1,99 @@
...
+        <Remirror manager={manager} initialContent={state} autoRender>
+          <ProsemirrorDevTools />
+        </Remirror>
+      </ThemeProvider>
+    </>
+  );
+};
+
+const html = String.raw; // Just for better editor support
+const content = html`<p>Drag and drop one or multiple files to the editor.</p>`;
```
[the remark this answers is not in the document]
ocavue: 🤔

packages/remirror__extension-file/src/file-loader.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * `FileLoader` is an interface based on the Web API `FileReader`. In remirror,
+ * the `FileReader` is used by default to read files uploaded by users, but if
+ * you don't want to use the browser's `FileReader`, you can design your own
+ * `FileReader`, as long as it implement this interface.
+ *
+ * Read more about the Web API `FileReader` here: https://developer.mozilla.org/en-US/docs/Web/API/FileReader
+ */
+export interface FileLoader {
```
whawker, line 9: I don't think this is needed anymore?

packages/remirror__extension-file/src/file-placeholder-plugin.ts
```diff
@@ -0,0 +1,71 @@
...
+      return { set: DecorationSet.empty, payloads: new Map<unknown, any>() };
+    },
+    apply(tr, { set, payloads }: PlaceholderPluginData): PlaceholderPluginData {
+      // Adjust decoration positions to changes made by the transaction
+      set = set.map(tr.mapping, tr.doc);
+      // See if the transaction adds or removes any placeholders
+      const action = tr.getMeta(this) as PlaceholderPluginAction | null;
+
+      if (action) {
+        const { add, remove } = action;
```
whawker, line 32: For consistency, I think we should use the same pattern as the [Annotation plugin](https://github.com/remirror/remirror/blob/beta/packages/remirror__extension-annotation/src/annotation-plugin.ts#L120-L142).

I.e. Instead of `action.add` or `action.remove` we have `action.type = ADD_PLACEHOLDER`/`action.type = REMOVE_PLACEHOLDER`

```ts
if (action.type === ADD_PLACEHOLDER) {
  // Do stuff
}

if (action.type === REMOVE_PLACEHOLDER) {
  // Do stuff
}
```

packages/remirror__extension-file/src/file-upload-context.ts
```diff
@@ -0,0 +1,43 @@
...
+  };
+
+  const set = (key: string, value: unknown) => {
+    values[key] = value;
+    afterSet();
+  };
+
+  const addListener = (listener: UploadContextListener) => {
+    eventListeners.push(listener);
+  };
```
whawker, line 25: We're using `nanoevents` [elsewhere in the codebase](https://github.com/remirror/remirror/blob/a2ca7a83f35b3831b97817eb2cb38b1a82d60ab8/packages/remirror__extension-positioner/src/positioner.ts#L299-L304), should we use that here too?

packages/remirror__extension-file/src/file-upload-context.ts
```diff
@@ -0,0 +1,43 @@
...
+    timeoutId = setTimeout(() => {
+      eventListeners.forEach((listener) => listener(values));
+      timeoutId = null;
+    });
+  };
+
+  return {
+    set,
+    get,
+    addListener: addListener,
```
whawker, line 41: Do we add a `removeListener` here too?

packages/remirror__extension-file/src/file-uploader.ts
```diff
@@ -0,0 +1,29 @@
...
+   * because of an error.
+   *
+   * `upload` can update the object `context` to update information during
+   * the upload process. `context` will be passed to the render function. The
+   * render function can add a listener to `context` by using `context.addListener`
+   * to get the updated values. The default render function will try to find the
+   * keys `loaded` and `total` in `context`, which are two numbers that
+   * represent the progress of the upload.
+   */
+  upload: (context: UploadContext) => Promise<FileAttributes>;
```
whawker, line 23: 🥂 excellent work here.

ocavue: Thanks

packages/remirror__extension-file/src/file-upload-context.ts
ocavue: Good idea. It does make the code simpler and cleaner.

packages/remirror__extension-file/src/file-upload-context.ts
ocavue: `addListener` will returns a callback that will remove the listener. Thanks to `nanoevents` this is so easy.

packages/remirror__extension-file/src/file-placeholder-plugin.ts
ocavue: done

packages/remirror__extension-file/src/file-component.tsx
```diff
@@ -0,0 +1,202 @@
...
+        padding: '8px 12px',
+        backgroundColor: '#e8ecf1',
+        color: '#000',
+        margin: '8px auto',
+        minHeight: '32px',
+        width: '100%',
+        maxWidth: '600px',
+        display: 'flex',
+        alignItems: 'center',
+      }}
```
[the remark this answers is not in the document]
ocavue: Added them to `@remirror/theme`.

whawker (approved):
Amazing work here!

ocavue:
Thank you very much for your help @whawker

PR #1054, opened 2021-08-11
docs: remove unnecessary corePreset pattern

ctbarna:
### Description

I noticed that the initialization of core presets was updated in code but not in the docs. This updates the docs with the new pattern. Please let me know if the formatting is correct.

### Checklist


ocavue:
@ctbarna Thank you for raise this. In fact, we don't need to add `corePreset` explicitly anymore. I'll also update some related docs in this PR.

PR #1069, opened 2021-08-18
feat(doc): allow docAttributes to have default values

whawker:
### Description

Doc extension currently supports a `docAttributes` option, however is only allows an array of keys, defaulting the value to `null`.

This PR allows `docAttributes` to be passed as a record to set the default/initial value of the attribute.

### Checklist

### DIFF
--- a/packages/remirror__extension-doc/src/doc-extension.ts
+++ b/packages/remirror__extension-doc/src/doc-extension.ts
@@ -3,8 +3,10 @@ import {
   command,
   CommandFunction,
   EditorSchema,
+  entries,
   extension,
   ExtensionPriority,
+  isPlainObject,
   NodeExtension,
   NodeExtensionSpec,
   NodeSpecOverride,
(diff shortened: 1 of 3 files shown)


.changeset/gold-fishes-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@remirror/extension-doc': patch
```
ocavue, line 2: let's also add `remirror` here

packages/remirror__extension-doc/src/doc-extension.ts
```diff
@@ -57,7 +59,7 @@ export interface DocOptions {
    *
    * The attributes are directly applied to the dom node.
```
ocavue, line 60: Maybe we need a little comment here. To explain what's the difference between array and object.

ocavue (approved):

Issue #1063, opened 2021-08-16
NodeFormattingExtension doesn't update attributes in DOM


ronnyroeller:
Reported by pepper via discord: https://discord.com/channels/726035064831344711/745695521305526302/875614371420725278

### Summary

Hi, I have reproduced a very minimal alignment example in codesandbox.
i.e. using the NodeFormattingExtension

post this,  i'm trying to select text and trying to run any of the leftAlign, centerAlign, rightAlign command.
But I don't see any change reflected on the editor.

I must be missing something obvious.

Any help is appreciated.

### Steps to reproduce

here's link to a minimal sandbox - https://codesandbox.io/s/remirror-basic-alignment-f392q?file=/src/Editor/Menu.jsx

ronnyroeller:
I can confirm that this is an issue. I created a branch with a storybook that shows the problem: https://github.com/remirror/remirror/tree/docs/node-formatting

The issue itself is caused because `attrs.style` doesn't actually contains the style, thereby the configured attributes are each applied to an empty string: https://github.com/remirror/remirror/blob/af894460b5521e0e18d67bcb42db245e83c85e6b/packages/remirror__extension-node-formatting/src/node-formatting-extension.ts#L232

I'm not sure how to solve this though. I can get the node from `options.node` but I'm not sure how I would get the `ApplySchemaAttributes` as this is a `PlainExtension`. See sketch below:
```diff
diff --git a/packages/remirror__extension-node-formatting/src/node-formatting-extension.ts b/packages/remirror__extension-node-formatting/src/node-formatting-extension.ts
index 9343d8cf7..2e5929da5 100644
--- a/packages/remirror__extension-node-formatting/src/node-formatting-extension.ts
+++ b/packages/remirror__extension-node-formatting/src/node-formatting-extension.ts
@@ -213,7 +213,7 @@ export class NodeFormattingExtension extends PlainExtension<NodeFormattingOption
           extractIndent(this.options.indents, element.style.marginLeft)
         );
       },
-      toDOM: (attrs) => {
+      toDOM: (attrs, options) => {
         // Ignoring the `0` value is intentional here.
         if (!attrs.nodeIndent) {
           return;
@@ -226,10 +226,12 @@ export class NodeFormattingExtension extends PlainExtension<NodeFormattingOption
           return;
         }
 
+        const { style: currentStyle = '' } = extra.dom(options.node);
+
         return {
           [NODE_INDENT_ATTRIBUTE]: indentIndex,
[...]

PR #1071, opened 2021-08-18
fix(extension-node-formatting): formatting wasn't reflected in style

ronnyroeller:
### Description

The SchemaAttributesObject didn't have access to the current style. This meant that only one of the formatting attributes was reflected in the style.

Solution provided by @whawker

Fixes https://github.com/remirror/remirror/issues/1063

### Checklist


whawker:
It's worth noting, that copy-and-paste doesn't work with the solution I provided - but I think that might be an existing issue with the NodeFormatting extension.

PR #1125, opened 2021-08-22
fix(i18n): set `sideEffect` as true because of global initialization

Tom910:
### Description

Right now with Webpack 5 i got error
```
Plurals for locale undefined aren't loaded. Use i18n.loadLocaleData method to load plurals for specific locale. Using other plural rule as a fallback.
```
It's happen because package @remirror/i18n run side effects in index.ts file and with sideEffects:true webpack 5 remove that code

### Checklist

### DIFF
--- a/.changeset/fifty-weeks-visit.md
+++ b/.changeset/fifty-weeks-visit.md
@@ -1,0 +1,5 @@
+---
+'@remirror/i18n': patch
+---
+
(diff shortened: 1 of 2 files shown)


ocavue:
@Tom910 Thank you. Released as `@remirror/i18n@1.0.3`

PR #1161, opened 2021-08-25
fix(core-types): extra attributes can be any JSON serializable value

whawker:
### Description

Previously only string values we allow, this PR utilises `type-fest` to allow for any JSON serializable value.

### Checklist


ocavue:
I didn't run into any use case so far, but I can imagine some: Let's say someone wants to build a form within remirror, then he may have a `multiSelect` node. This node could have two attributes: `{ options: string[], selectedIndex: number }` 

If we won't allow array and object here, we might get into trouble in the future. 

To make the implementation simpler, could we just use `any`, like what prosemirror dose? What TS can do here is already limit anyway, since we won't use generic to give different types to different nodes.

PR #1207, opened 2021-09-02
fix: remove react dependency from wysiwyg-preset

ronnyroeller:
### Description

Commit https://github.com/remirror/remirror/commit/9b8f7662775fa8228b43c327d497ccbf30c0c730 had introduced a dependency from wysiwyg to the react-part of remirror. This removes this dependency again.

### Checklist

### DIFF
--- a/.changeset/calm-olives-repair.md
+++ b/.changeset/calm-olives-repair.md
@@ -1,0 +1,5 @@
+---
+'@remirror/preset-wysiwyg': minor
+---
+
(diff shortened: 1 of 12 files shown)


ronnyroeller:
@ocavue Does this address your concern regarding the React dependency?

ocavue:
> @ocavue Does this address your concern regarding the React dependency?

Yep. I pushed a script to make sure that `remirror` doesn't depend on `react`.

Maybe we should add a changeset so that users won't be surprised if they already use table with wysiwyg-preset.

PR #1208, opened 2021-09-02
fix(core): unchained commands should use a new transaction

### LINKED ISSUE #1197: TaskList toggle button crashes editor when cursor is placed in task list
### Steps to reproduce

1. Open editor with toggle menu button for `toggleTaskList`, e.g. https://remirror.vercel.app/?path=/story/editors-social--basic
2. Add a task list, e.g. via toggle menu button (or enter `[ ]`)

The same happens when adding a bullet or ordered list node to the document. It doesn't happen if there is no toggle button for `toggleTaskList`.

https://user-images.githubusercontent.com/9339055/131669582-eaa8e48d-1c47-4d1b-a9cf-3d3e92fe9715.mov

### Expected results

Task list is added

### Actual results

Editor crashes

image

### Possible Solution

The error is caused by `list-commands.deepChangeListType`: https://github.com/remirror/remirror/blob/4d5b617b29a30bf3728b4e3ab738db2e7b244d19/packages/remirror__extension-list/src/list-commands.ts#L347

I suspect the issue is that the toggle menu button calls `toggleTaskList` to check if it could be executed (i.e. to disable the button or not) but the code actually tries to change the document.

I don't really understand why this happens only for task lists but not for bullet or ordered lists.

whawker:
### Description

Unchained commands should use a new transaction to prevent leaking of previous command steps

Fixes #1197

### Checklist

### DIFF
--- a/.changeset/twelve-hotels-matter.md
+++ b/.changeset/twelve-hotels-matter.md
@@ -1,0 +1,5 @@
+---
+'@remirror/core': patch
+---
+
+Unchained commands should use a new transaction to prevent leaking of previous command steps
--- a/packages/remirror__core/__tests__/remirror-manager.spec.tsx
+++ b/packages/remirror__core/__tests__/remirror-manager.spec.tsx
@@ -17,7 +17,7 @@ import type {
   ProsemirrorAttributes,
 } from '@remirror/core-types';
 import { Schema } from '@remirror/pm/model';
-import { EditorState, Plugin } from '@remirror/pm/state';
+import { EditorState, Plugin, Transaction } from '@remirror/pm/state';
 import { EditorView } from '@remirror/pm/view';
 
 describe('Manager', () => {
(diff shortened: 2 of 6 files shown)


ocavue:
@whawker  All two tests can be fixed by making the editor from a controlled editor to an uncontrolled editor. I don't understand the life cycle difference between these two kinds of editors but I can reproduce the "commands are unable to chain in a controlled editor" with storybook. Check https://pr1208-remirror-ocavue.vercel.app/?path=/story/extensions-italic--basic for a demo.

ocavue:
It doesn't feel right that a controlled editor couldn't call two commands at once.

whawker:
Seems this is expected actually, see the [second paragraph under Potential pitfalls here](https://remirror.io/docs/react/controlled/#potential-pitfalls)

ocavue:
Oh, I see. So it's a known issue. I think we can continue this patch, with some warning in the changeset/changelog: this patch fixes a bug that could bring breaking change if users are using a controlled editor and calling multiple commands at once.

whawker:
Seems fine to me. I wouldn't consider it a breaking change if it's aligning with how the docs _say_ it works.

Definitely worth highlighting it though. Just in case.

whawker:
Released with warning, as previously discussed.

https://github.com/remirror/remirror/releases/tag/remirror%401.0.22

Issue #1230, opened 2021-09-10
Link extension detects parts of emails


ronnyroeller:
### Summary

Entering an email address with a dot makes the link extension detect the username as a link.

### Steps to reproduce

1. Open storybook: https://remirror.vercel.app/?path=/story/extensions-link--basic
2. Enter `first.last@example.com`

### Expected results

Email address is not detected as link (alternatively: complete email address is detected)

### Actual results

Only first part `first.last` is detected as link:
![image](https://user-images.githubusercontent.com/9339055/132867140-bd732e8f-2d3d-4123-b39d-d2700efdd077.png)

### Possible Solution

The RegEx could require a whitespace (`\s`) for matching and then ignore it afterwards as part of the detected link: https://github.com/remirror/remirror/blob/main/packages/remirror__extension-link/src/link-extension.ts#L175

PR #1232, opened 2021-09-13
fix: prevent to detect parts of emails as link

ronnyroeller:
### Description

Fixes https://github.com/remirror/remirror/issues/1230

### Checklist

- [ x] I have updated the documentation where necessary.


packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -669,6 +671,13 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
  * Extract the `href` from the provided text.
  */
 function extractHref(url: string, defaultProtocol: DefaultProtocol) {
+  // This isn't 100% precise because we allowed URLs without protocol
+  // For example, userid@example.com could be email address or link http://userid@example.com
+  const isEmail = url.includes('@');
```
whawker, line 676: Feel like we should check there is no protocol present on the URL?

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -172,7 +172,9 @@ export type LinkAttributes = ProsemirrorAttributes<{
     defaultProtocol: '',
     selectTextOnClick: false,
     openLinkOnClick: false,
-    autoLinkRegex: /(?:https?:\/\/)?[\da-z]+(?:[.-][\da-z]+)*\.[a-z]{2,8}(?::\d{1,5})?(?:\/\S*)?/gi,
+    // Based on https://gist.github.com/dperini/729294
+    autoLinkRegex:
+      /(?:(?:(?:https?|ftp):)?\/\/)?(?:\S+(?::\S*)?@)?(?:(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z0-9\u00a1-\uffff][a-z0-9\u00a1-\uffff_-]{0,62})?[a-z0-9\u00a1-\uffff]\.)+(?:[a-z\u00a1-\uffff]{2,}\.?))(?::\d{2,5})?(?:[/?#]\S*)?/gi,
```
whawker, line 177: Should also update the comment on line 120

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -172,7 +172,9 @@ export type LinkAttributes = ProsemirrorAttributes<{
     defaultProtocol: '',
     selectTextOnClick: false,
     openLinkOnClick: false,
-    autoLinkRegex: /(?:https?:\/\/)?[\da-z]+(?:[.-][\da-z]+)*\.[a-z]{2,8}(?::\d{1,5})?(?:\/\S*)?/gi,
+    // Based on https://gist.github.com/dperini/729294
+    autoLinkRegex:
+      /(?:(?:(?:https?|ftp):)?\/\/)?(?:\S+(?::\S*)?@)?(?:(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[01])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4])|(?:(?:[\da-z\u00A1-\uFFFF][\w\u00A1-\uFFFF-]{0,62})?[\da-z\u00A1-\uFFFF]\.)+[a-z\u00A1-\uFFFF]{2,}\.?)(?::\d{2,5})?(?:[#/?]\S*)?/gi,
```
whawker, line 177: Wondering fi we remove this part, do we need to catch the `mailto:` case?

```suggestion
      /(?:(?:(?:https?|ftp):)?\/\/)?(?:(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[01])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4])|(?:(?:[\da-z\u00A1-\uFFFF][\w\u00A1-\uFFFF-]{0,62})?[\da-z\u00A1-\uFFFF]\.)+[a-z\u00A1-\uFFFF]{2,}\.?)(?::\d{2,5})?(?:[#/?]\S*)?/gi,
```

ronnyroeller: I think this would bring us back to the original issue that `first.last@example.com` would detect only `first.last`.

whawker: You're right 👍

whawker: Feel like we should remove the IP address section? Making the protocol optional seems to make it not match `127.0.0.1`, it matches `27.0.0.1` instead?

PR #1257, opened 2021-09-22
docs: embed generated api document

ocavue:
### Description

1. Remove `API` in the navbar and move all the content under the `Docs`
2. Add the new `API` section in the sidebar under the `Docs`, which contains all the document generated from the TypeScript source code. For now, there are only three packages (`@remirror/extension-strike`, `prosemirror-suggest`, `prosemirror-trailing-node`) and I'll include most (if not all) packages in other pull request. 

Check the comment posted by github-actions below to see a live site.

Screenshot 2021-09-22 at 23 58 13

### Checklist

### Future works

There are something need to improved in the future:

- There are some broken links in the generated API document (e.g. `enum` seems doesn't work well)
- Some marks we wrote inside the comment don't fit the requirement from `@microsoft/api-extractor` (e.g. we used `@default` but `@microsoft/api-extractor` requires `@defaultValue`)
- We might somehow combine the existed handwritten extension document and the new generated extension document. For example, we write document in the TypeScript comment instead of in a markdown file, and then we generate the markdown file. This could make the sidebar more clear.


ronnyroeller:
> We might somehow combine the existed handwritten extension document and the new generated extension document. For example, we write document in the TypeScript comment instead of in a markdown file, and then we generate the markdown file. This could make the sidebar more clear.

I like the idea of adding them as TypeScript comment. Most of the handwritten docu is a copy of the TypeScript comments anyway. So, we could go through the existing docu, and just copy the relevant parts back to TypeScript comments. I'd be happy to help with that.

ocavue:
I will look into it and try to find a good way to implement it.

PR #1324, opened 2021-10-17
feat(core): disable persistent selection by default

### LINKED ISSUE #893: Text selection will change the color after mouse up
### Summary

When using Safari, the text selection color will change after the mouse up.
When using Chrome, not only the text selection color change, there is also two text selection overlay after the mouse up.

### Steps to reproduce

1. checkout to the beta branch 
2. run `pnpm storybook`
3. open http://localhost:6006/?path=/story/link-extension--basic
4. select some text
  

### Screenshot(s)
 

https://user-images.githubusercontent.com/24715727/111064280-0bcd0300-84ee-11eb-939b-1dadfb7f9ea2.mov

### LINKED ISSUE #1282: Proposal: deprecate `persistentSelectionClass`
### Description

[`persistentSelectionClass`](https://github.com/remirror/remirror/blob/270edd91ba6badf9468721e35fa0ddc6a21c6dd2/packages/remirror__core/src/builtins/decorations-extension.ts#L25-L31) is a setting for adding a decoration to the selected text and can be used to preserve the marker for the selection when the editor loses focus. This causes some problems:

- First of all, we don't need it since the browser can already preserve the selection (and change its background color to gray) when the editor loses focus.

  - If we don't want the selected text to become gray when we click a button on the menu and blur the editor, we can use `chainedCommands` to focus the editor again. That's how ProseMirror's official example does.

      ```diff
       const BoldButton = () => {
      -  const commands = useCommands();
      +  const chain = useChainedCommands();
         const active = useActive(true);
         return (
      -     <button onClick={() => commands.toggleBold()} className={cx(active.bold() && 'active')}>
      +     <button onClick={() => chain.toggleBold().focus().run()} className={cx(active.bold() && 'active')}>
             Bold
           </button>
         );
       };
      ```

- This causes different and unexpected behaviors on browsers: #893

- This causes poor performance when the selection contains a lot of text (You can see that the selection changed when I click the editor)

   https://user-images.githubusercontent.com/24715727/135662930-36da33f3-29ae-4af5-9a10-dda499b7b168.mp4

So I propose that we

- Change the default value of `persistentSelectionClass` from `selection` to `undefined`, so the editor won't add selection decoration by default. 

- Deprecate this setting remove it on v2. If someone really wants this back, they only need to copy [this method](https://github.com/remirror/remirror/blob/270edd91ba6badf9468721e35fa0ddc6a21c6dd2/packages/remirror__core/src/builtins/decorations-extension.ts#L326-L340) to their only editor, which is about 10 lines of code.

ocavue:
### Description

 
Closes #1282
Closes #893 

1. disable persistent selection by default (*breaking change*). 
2. only show persistent selection if the editor loses focus. 

Check the discussion at #1282 for more details. 

### Checklist

### DIFF
--- a/.changeset/lovely-jars-live.md
+++ b/.changeset/lovely-jars-live.md
@@ -2,7 +2,7 @@
 '@remirror/core': minor
 ---
 
-**BREAKING CHANGE**: The option `persistentSelectionClass` for `DecorationsExtension` is now `undefined` by default. It needs to be explicitly configured as `'selection'` to enable persistent selection.
+**BREAKING CHANGE**: The option `persistentSelectionClass` for `DecorationsExtension` is now `undefined` by default. It needs to be explicitly configured to enable persistent selection. You can set it as `'selection'` to match the default styles provided by `@remirror/styles`.
 
 If you are using `@remirror/react`, you can enable it like this:
 


ocavue:
@ankon @whawker Could you review this if you get some time?

ocavue:
@whawker  First of all, thank you so much for the review. Much appreciated.


> but (for the moment) you are required to explicitly opt out?

I saw [this comment](https://github.com/remirror/remirror/issues/1282#issuecomment-936360984) so I decided to disable it by default. Considering that the impact of this breaking change is sufficiently small, disabling it shouldn't be a big issue.

PR #1423, opened 2021-12-10
feat: support shortcut for left/right arrows

ronnyroeller:
### Description

Allow to replace `<->` with a more pretty left/right arrow

### Checklist

### Screenshots

image

### DIFF
--- a/.changeset/polite-spoons-stare.md
+++ b/.changeset/polite-spoons-stare.md
@@ -1,0 +1,5 @@
+---
+'@remirror/extension-shortcuts': minor
+---
+
+feat: support shortcut for left/right arrows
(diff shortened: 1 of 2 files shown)


ankon:
I would have expected the replacement to be ['LEFT RIGHT ARROW' (U+2194)](https://www.fileformat.info/info/unicode/char/2194/index.htm) to be consistent with the other arrows, but you chose ['LEFT RIGHT BLACK ARROW' (U+2B0C)](https://www.fileformat.info/info/unicode/char/2b0c/index.htm).

We discussed this in a call earlier, this is a reminder to document the choice here indeed as well :)

ronnyroeller:
I picked the Black Arrow because it looked more visually balanced. Yet, checking again I see that I compared it to the wrong left/right arrow. I'll create another PR to change to `LEFT RIGHT ARROW (U+2194)`.

PR #1451, opened 2021-12-31
chore: prefix `main` and `module` fields with `./`

ilkkao:
### Description

Esbuild currently ignores the browser field in the core utilies package, causing jsdom inclusion which leads to errors (esbuild doesn't show the call stack so the snippet is not that informative) :

```
$ ts-node ./esbuild.ts
DEV mode: false
✘ [ERROR] Could not resolve "path"

    node_modules/jsdom/lib/api.js:2:21:
      2 │ const path = require("path");
        ╵                      ~~~~~~

  The package "path" wasn't found on the file system but is built into node. Are you trying to
  bundle for node? You can use "platform: 'node'" to do that, which will remove this error.
```

The file names in the mapping need to match exactly. Therefore I used the ./ notation for the main and module entries. This is in line how the other package internal files are references in the `package.json`.

### Checklist

- [X] I have read the [**contributing**](https://github.com/remirror/remirror/blob/HEAD/docs/contributing.md) document.


ilkkao:
here is how esbuild interprets `main`, `module`, and `browser` fields in package.json https://esbuild.github.io/api/#main-fields

For `browser` field it links to the spec here: https://gist.github.com/defunctzombie/4339901/49493836fb873ddaa4b8a7aa0ef2352119f69211

There the examples show that a path that doesn't start with `./` refers to separate module. Looks like esbuild strictly follows the spec.

There are many other places that use `dist...` instead of `./dist` in `main` and `module` fields. I guess those should be updated too?

ocavue:
Hi. Thanks for your first PR ❤️

In addition to the esbuild link above, I also checked the documentation for some other tools. It seems that all tools agree that it's a good practice to add `./` at the beginning of `main` and `module` fields, so let's do it. 

Node.js docs https://nodejs.org/docs/latest-v16.x/api/packages.html#main-entry-point-export

> #### Main entry point export[#](https://nodejs.org/docs/latest-v16.x/api/packages.html#main-entry-point-export)
> 
> To set the main entry point for a package, it is advisable to define both [`"exports"`](https://nodejs.org/docs/latest-v16.x/api/packages.html#exports) and [`"main"`](https://nodejs.org/docs/latest-v16.x/api/packages.html#main) in the package’s [`package.json`](https://nodejs.org/docs/latest-v16.x/api/packages.html#nodejs-packagejson-field-definitions) file:
> 
> ```json
> {
>   "main": "./main.js",
>   "exports": "./main.js"
> }
> ```

esbuild https://esbuild.github.io/api/#main-fields


> For package authors: If you want to author a package that uses the browser field in combination with the module field to fill out all four entries in the full CommonJS-vs-ESM and browser-vs-node compatibility matrix, you want to use the expanded form of the browser field that is a map instead of just a string:
>
> ```json
> {
>   "main": "./node-cjs.js",
>   "module": "./node-esm.js",
>   "browser": {
>     "./node-cjs.js": "./browser-cjs.js",
>     "./node-esm.js": "./browser-esm.js"
>   }
> }
> ```

Vite https://vitejs.dev/guide/build.html#library-mode

> Recommended `package.json` for your lib:
>
> ```
> {
>   "name": "my-lib",
>   "files": ["dist"],
>   "main": "./dist/my-lib.umd.js",
>   "module": "./dist/my-lib.es.js",
>   "exports": {
>     ".": {
>       "import": "./dist/my-lib.es.js",
>       "require": "./dist/my-lib.umd.js"
>     }
>   }
> }
> ```





We have a script to generate these fields in `package.json`. I will edit this script and release a new version for all NPM packages.

ocavue:
Preconstruct [doesn't like](https://github.com/preconstruct/preconstruct/issues/437) `./`. I will disable Preconstruct for now.

ilkkao:
@ocavue great, makes sense of course that the fields are autogenerated. I won't touch this PR anymore. In the unlikely case a workaround would be needed for esbuild, I'e used `yarn install && rm -fr node_modules/jsdom`. Jsdom is required inside a try/catch, so optional.

btw just switched from slate in my project and feel so much more productive now.

ocavue:
@ilkkao New version has been released. You might need to remove `node_modules` or even `yarn.lock` to update remirror to the latest version correctly.

Issue #772, opened 2020-11-11
History extension overrides keymap from yjs extension


tommoor:
### Summary

- If using the yjs extension the key bindings for undo/redo are overridden by the history extension.

### Steps to reproduce

- Use the yjs extension
- Type something
- Use `cmd+z` to undo
- Use `cmd+shift+z` to redo

### Expected results

- Change is redone.

### Actual results

- Change is not redone

### Possible Solution

- Return a tuple from `createKeymap` in the yjs extension to indicate that the keymapping there should take priority?

tommoor:
Heads up: https://github.com/remirror/remirror/actions/runs/356864978

ifiokjr:
Thanks for the heads up. I think that's been resolved by this PR https://github.com/JasonEtco/is-sponsor-label-action/pull/15 now.

ifiokjr:
I'm trying to replicate this and can't. By default, the `YjsExtension` has a higher priority (`ExtensionPriority.High`) so when included it should automatically take precedence.

 Is it possible that the history extension or `CorePreset` has been given a higher priority in your editor setup?

I'm creating a PR which gives the `yjs` key bindings a priority of `High` in case that helps.

tommoor:
>  Is it possible that the history extension or CorePreset has been given a higher priority in your editor setup?

Not in this case.

As I understand it the keybindings have their own separate priority stack?

tommoor:
Interestingly `CMD+Y` works, but `CMD+SHIFT+Z` does not – so that's an additional clue!

ifiokjr:
> Interestingly CMD+Y works, but CMD+SHIFT+Z does not – so that's an additional clue!

Ahhh, amazing. I think that's the issue.

PR #1466, opened 2022-01-12
docs: include the note that the `history` extension conflicts with `yjs`

ankon:
### Description

The `yjs` and `history` extensions conflict, for example by both providing a `undo` command: In a yjs-enabled remirror instance with the `history` extension not excluded from the core preset the `undo` command will actually be provided the `history` extension, which isn't aware of the specific Yjs transactions -- and for instance will allow to "undo" the initial content of the editor.

This is mentioned in the JS docs for the `excludeExtensions` setting: https://github.com/remirror/remirror/blob/64593d99c3d88efe2b46a990b873e0ef107d01dd/packages/remirror__preset-core/src/core-preset.ts#L25-L26

This PR brings that important tidbit into a place that a user might actually see: The Yjs extension documentation :)

While there: Drop some weird half-sentences in the same document. I don't know where the writer wanted to go there, but the basic example seems to be good enough to show how to use the extension and get going.

### Checklist

- [x] ~New code is unit tested and all current tests pass when running `pnpm test`.~ No new code.

### DIFF
--- a/docs/extensions/yjs-extension.mdx
+++ b/docs/extensions/yjs-extension.mdx
@@ -23,9 +23,16 @@ You can use the imports in the following way.
 import { YjsExtension } from 'remirror/extensions';
 ```
 
-To install it directly you can use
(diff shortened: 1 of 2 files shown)


ankon:
Historic "related" issue: https://github.com/remirror/remirror/issues/772

PR #1488, opened 2022-01-28
fix: avoid double processing of annotations

ankon:
### Description

When an action is given we anyways refresh the cached annotations from the
backing store, so there is no point to first map all of them through the transform.

Note that actions are actually rare, so in absolute numbers this likely doesn't "speed up" things.

### Checklist


ronnyroeller:
@ankon Could you pls resolve the conflict, and then I can merge?

whawker (approved):

PR #1490, opened 2022-01-28
fix: update annotations in Yjs transaction

ankon:
### Description

Annotations with the Yjs extension enabled are stored in a Y.Map, which is "almost but not quite" a Map. But, apart from that, the pattern in which is should be used is quite different, and changes that
conceptually belong together should be executed inside a single transaction.

This PR has two parts:
1. Refactor how the annotation extension stores the annotations, and instead of exposing a "map + transform" interface expose a more flexible "store" interface.
2. Use that interface and implement `setAnnotations` and `removeAnnotations` for the yjs extension with a `Y.transact` wrapper

Note that I would have loved to kill the `getMap` part, because there shouldn't be any other user of this (it feels like a "escape hatch" purely intended for the Yjs extension), but semver-wise that would be problematic. So, the code does have backwards compatibility for it, but it will throw an AssertionError in non-production environments (`getStore` has a non-`undefined` default, and providing both `getStore` and `getMap` is not allowed)

### Checklist

### DIFF
--- a/packages/remirror__extension-annotation/__tests__/annotation-extension.spec.ts
+++ b/packages/remirror__extension-annotation/__tests__/annotation-extension.spec.ts
@@ -2,6 +2,7 @@ import { extensionValidityTest, renderEditor } from 'jest-remirror';
 
 import type { AnnotationOptions } from '../';
 import { Annotation, AnnotationExtension } from '../';
+import { MapLikeAnnotationStore } from '../src/annotation-store';
 
 extensionValidityTest(AnnotationExtension);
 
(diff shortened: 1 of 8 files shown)


ronnyroeller:
Thanks, @ankon . I merged the PR.
I just realized that the PR didn't have a changeset, ie it will only roll out with the next regular release.
In case you require a release with the PR: could you please create another PR with only a changeset (`pnpm changeset`)?

ankon:
Thanks. Closing open ends: No, this was intentional, following the regular release cadence should be fine -- but if I end up depending on this change and it hasn't been released then I'll create a changeset PR indeed.

PR #1496, opened 2022-01-31
refactor(extension-link): refactor auto link to simplify and remove bugs

whawker:
### Description

Currently auto linking is performed by a suggestor - however auto linking it not trigger by a specific character (like `@` for instance), this means when are abusing (and constrained) by the suggestor API making this area of the code particularly difficult to maintain.

This PR modifies how auto linking works, it uses changed ranges to determine auto link candidates to ensure good performance on larger documents.

### Checklist

#### One of the auto link issues resolved in this PR

https://user-images.githubusercontent.com/2003804/151844465-5848b99b-b78b-441a-8ede-6cd97209b8ad.mov

### DIFF
--- a/packages/remirror__extension-link/__tests__/link-extension.spec.ts
+++ b/packages/remirror__extension-link/__tests__/link-extension.spec.ts
@@ -517,7 +517,7 @@ describe('autolinking', () => {
     );
   });
 
-  it('can parse telphone by overriding extractHref', () => {
+  it('can parse telephone by overriding extractHref', () => {
     const phoneRegex = /(?:\+?(\d{1,3}))?[(.-]*(\d{3})[).-]*(\d{3})[.-]*(\d{4})(?: *x(\d+))?/;
     const linkRegex =
       /(?:(?:(?:https?|ftp):)?\/\/)?(?:\S+(?::\S*)?@)?(?:(?:[\da-z\u00A1-\uFFFF][\w\u00A1-\uFFFF-]{0,62})?[\da-z\u00A1-\uFFFF]\.)+[a-z\u00A1-\uFFFF]{2,}\.?(?::\d{2,5})?(?:[#/?]\S*)?/gi;
@@ -602,7 +602,6 @@ describe('autolinking', () => {
       range: {
         from: 1,
         to: editorText.length + 1,
-        cursor: editorText.length + 1,
       },
       attrs: { auto: true, href: '//test.co' },
     });
@@ -616,7 +615,6 @@ describe('autolinking', () => {
       range: {
         from: 1,
         to: editorText.length + 1,
-        cursor: editorText.length + 1,
       },
       attrs: { auto: true, href: '//test.com' },
     });
@@ -656,6 +654,18 @@ describe('autolinking', () => {
     );
   });
 
+  it('updates the autolink if modified in the middle', () => {
+    editor
+      .add(doc(p('<cursor>')))
+      .insertText('github.com')
+      .selectText(7)
+      .insertText('status');
+
+    expect(editor.doc).toEqualRemirrorDocument(
+      doc(p(link({ auto: true, href: '//githubstatus.com' })('githubstatus.com'))),
+    );
+  });
+
   it('supports deleting selected to to invalidate the match', () => {
     editor
       .add(doc(p('<cursor>')))
@@ -673,6 +683,14 @@ describe('autolinking', () => {
       .backspace(2);
 
     expect(editor.doc).toEqualRemirrorDocument(doc(p('test.c')));
+  });
+
+  it('supports backspace to create a match', () => {
+    editor.add(doc(p('test. <cursor>com'))).backspace();
+
+    expect(editor.doc).toEqualRemirrorDocument(
+      doc(p(link({ auto: true, href: '//test.com' })('test.com'))),
+    );
   });
 
   it('can respond to `Enter` key presses', () => {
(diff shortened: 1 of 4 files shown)


packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -267,18 +283,13 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
-        this.store.removeSuggester(this.name);
-      }
-    }
+  protected init() {
+    super.init();
+    const { autoLinkRegex } = this.options;
+    this._autoLinkRegexNonGlobal = new RegExp(
+      `^{autoLinkRegex.source}$`,
+      autoLinkRegex.flags.replace('g', ''),
+    );
```
whawker, line 292: This is icky 🤢 , but it is either this or reseting `this.autoLinkRegex.lastIndex = 0` after each match

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -371,224 +382,6 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
-      {
-        type: 'mark',
-        regexp: this.options.autoLinkRegex,
-        markType: this.type,
-        getAttributes: (url, isReplacement) => ({
-          href: getMatchString(url),
-          auto: !isReplacement,
-        }),
-      },
-    ];
```
whawker, line 388: Paste rule was always applied, even when `this.options.autoLink = false` - is this expected behaviour?

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -237,9 +255,7 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
             }
 
             const href = node.getAttribute('href');
-            const auto =
-              node.hasAttribute(AUTO_ATTRIBUTE) ||
-              this.options.autoLinkRegex.test(node.textContent ?? '');
+            const auto = node.hasAttribute(AUTO_ATTRIBUTE);
```
whawker, line 258: This is parsing an anchor tag already, we shouldn't indicate it is an auto link in my opinion?

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -205,6 +210,15 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
   }
 
+  /**
+   * The autoLinkRegex option with the global flag removed, ensure no "lastIndex" state is maintained over multiple matches
+   * @private
+   */
+  private _autoLinkRegexNonGlobal = new RegExp(
+    `^${DEFAULT_AUTO_LINK_REGEX.source}$`,
+    DEFAULT_AUTO_LINK_REGEX.flags.replace('g', ''),
+  );
```
ankon, line 220: Is this needed here, if `init` will anyways update it? The replace might be "icky", but if it is in just one place it could be documented there to stand a chance of survival :)

whawker: Yep fair point, just added it to avoid a `this. _autoLinkRegexNonGlobal` might be undefined type warning.

packages/remirror__extension-link/__tests__/link-extension.spec.ts
```diff
@@ -517,7 +517,7 @@ describe('autolinking', () => {
     );
   });
 
-  it('can parse telphone by overriding extractHref', () => {
+  it('can parse telephone by overriding extractHref', () => {
```
ocavue, line 520: Thanks for fixing it. Can't imagine how many typos I've left in the codebase. 😭

whawker:
Any more thoughts @ocavue?

ocavue:
Seems that editing the text inside the link will not update the href: 

https://user-images.githubusercontent.com/24715727/152427798-1335d2a1-c998-4610-ad3c-fafcb211b5ae.mp4

whawker:
☹️ good spot!

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -693,6 +563,66 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
+    }
+
+    return linkMarks;
+  }
+
+  private findAutoLinks(str: string): FoundAutoLink[] {
+    const toAutoLink: FoundAutoLink[] = [];
+    let match;
+
+    while ((match = this.options.autoLinkRegex.exec(str)) !== null) {
```
ocavue, line 609:
```suggestion
    for (const match of findMatches(str, this.options.autoLinkRegex)) {
```

Maybe we can use the `findMatches` function from `@remirror/core` here. It handles the global flag and lastIndex for us.

packages/remirror__extension-link/src/link-extension.ts
ocavue: I believe it's expected. For example, Google Docs doesn't have auto-linking (you need to parse the space to create a link) but it does have a paste rule for links.

packages/remirror__extension-link/src/link-extension.ts
ocavue: It might not a big difference here. I'll go with your choice.

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -683,6 +465,94 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
+            this.findAutoLinks(nodeText)
+              // calculate link position
+              .map((link) => ({
+                ...link,
+                from: pos + link.start + 1,
+                to: pos + link.end + 1,
+              }))
+              .filter((link) => {
+                // Determine if found link is within the changed range
+                return within(link.from, from, to) || within(link.to, from, to);
```
ocavue, line 530: @whawker  Seems that "editing inside a link won't update href" issue comes from here

whawker:
> Seems that editing the text inside the link will not update the href:

Fixed, and [test added](https://github.com/remirror/remirror/blob/e5eef6dfc6536a7227a7e5dee60ef8a7403efbb3/packages/remirror__extension-link/__tests__/link-extension.spec.ts#L657-L667)

ocavue (approved):

PR #1502, opened 2022-02-02
fix: replace "assert" with defensive "if"

### LINKED ISSUE #1501: Annotation extension race between `updateAnnotation` and concurrent editors (yjs)
### Summary

`updateAnnotations` asserts that the provided annotation is in the map of annotations. This assertion is wrong if the map is actually a Yjs map, and the caller is talking about an older version of the annotation while the map has already been updated by a concurrent editor.

There is a local variant of this problem with just a single editor: If you trigger a call to `formatAnnotations` when the map contains a annotation with `!from || !to`, and then iterate over the results of the `getAnnotations` helper, you will actually step over an annotation that has been deleted. This was fixed as a side-effect of factoring out an annotation store and not deleting the potentially invisible annotations at this point, so here a historic link to the problem (note the `map.delete` followed by `annotations.push`):

https://github.com/remirror/remirror/blob/2a9c611edf335df515f8edd007ca98121ae9ec31/packages/remirror__extension-annotation/src/annotation-plugin.ts#L81-L100

### Possible Solution

* Remove the assert in `updateAnnotations` and replace with a defensive `if`
* Release changes of https://github.com/remirror/remirror/pull/1490

ankon:
### Description

Remove assert.

Fixes #1501.

### Checklist

### DIFF
--- a/.changeset/serious-baboons-hear.md
+++ b/.changeset/serious-baboons-hear.md
@@ -1,0 +1,6 @@
+---
+'@remirror/extension-annotation': patch
+'@remirror/extension-yjs': patch
+---
+
+Avoid race conditions in processing annotations
--- a/packages/remirror__extension-annotation/src/annotation-store.ts
+++ b/packages/remirror__extension-annotation/src/annotation-store.ts
@@ -1,5 +1,3 @@
-import { assert } from '@remirror/core';
-
 import { Annotation } from '.';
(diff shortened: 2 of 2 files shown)


ankon:
Link to #1490: Changeset is here :)

ocavue (approved):

PR #1512, opened 2022-02-07
feat: allow to disable yjs undo/redo

ankon:
### Description

The yjs extension provides a undo/redo implementation, which the docs suggest should be used instead of the default one provided in remirror (through the history extension). This suggestion is somewhat problematic, depending on the use case:
1. `yUndo` is using the Yjs `UndoManager`, which works on the level of the Yjs document, and as such actually collects also changes of other users. The manager is configured to watch transactions from the sync-plugin.
2. Even worse (IMHO) is that while the undo manager is able to track changes in a very fine-grained way ("character appeared over here"), the yjs extension applies any such changes by rendering the complete prosemirror document based on the Yjs document, and then simply replacing the document inside the editor with that. That makes the document change appear to be completely different to any other place that tracks positions, such as the annotation information.

In our environment we found that both of these behaviors are problematic: We need annotations, and from a user's point of view it is really odd to press your undo key, and see a change disappear from another currently active user. 

Unfortunately the fix for this is requiring multiple parts:
* remirror (this PR): Make it possible to disable the yjs extension `yUndo`/`yRedo` commands (and, as a side-effect, save the memory for the undo manager tracking changes)
* y-prosemirror (https://github.com/yjs/y-prosemirror/pull/91): Mark all transactions from the sync-plugin to be not included in the changes the history extension tracks for undo purposes

**Note**: I think it is fine to merge this change without https://github.com/yjs/y-prosemirror/pull/91, as disabling the UndoManager is already quite something. But, a note might be helpful in the docs that one needs a modified `y-prosemirror` library as well. In our case we actually have [that](https://github.com/Collaborne/y-prosemirror) anyways for historic reasons, *and* a [@remirror/extension-yjs fork](https://github.com/Collaborne/remirror-extension-yjs) using that library. While both of these are open-source, we do _NOT_ recommend to use them in your own projects though, as they receive fairly experimental changes whenever we need them. :)

### Checklist

### DIFF
--- a/.changeset/dry-jars-relax.md
+++ b/.changeset/dry-jars-relax.md
@@ -1,0 +1,5 @@
+---
+'@remirror/extension-yjs': patch
+---
+
+Allow disabling the yjs undo functionality in the yjs extension
--- a/docs/extensions/yjs-extension.mdx
+++ b/docs/extensions/yjs-extension.mdx
@@ -23,7 +23,9 @@ You can use the imports in the following way:
 import { YjsExtension } from 'remirror/extensions';
 ```
 
-The `yjs` extension provides support for "undo"/"redo" commands, which conflicts with the default `history` extension. You should therefore disable the `history` extension when using the `yjs` extension in the core preset configuration:
+The `yjs` extension provides support for "undo"/"redo" commands, which conflicts with the default `history` extension. The `history` extension provides "undo"/"redo" for the changes done by the _current user_ only, while the `yjs` extension provides "undo"/"redo" on the level of the underlying yjs transactions, which covers changes by _all active users._ Note however that a `yjs` "undo" replaces the complete document, which prevents other extensions (such as the `annotations` extension) from tracking positions.
+
+You can select the `yjs` "undo"/"redo" implementation by disabling the `history` extension in the core preset configuration:
 
 ```ts
 const { manager } = useRemirror({
@@ -33,6 +35,16 @@ const { manager } = useRemirror({
   },
 });
 ```
+
+Alternatively you can also disable the "undo"/"redo" functionality of the `yjs` extension:
+
+```ts
+const { manager } = useRemirror({
+  extensions: () => [new YjsExtension({ getProvider, disableUndo: true })],
+});
+```
+
+_Note that using the `history` extension "undo"/"redo" requires [additional support from the `y-prosemirror` library](https://github.com/yjs/y-prosemirror/pull/91)._
 
 ### Examples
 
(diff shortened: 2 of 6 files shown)


whawker:
Worth adding a warning in the YJS extension when the annotation extension is present, but the `disableUndo` option is not set?

Possibly [here](https://github.com/ankon/remirror/blob/e926ff64f66aee6d1510ae1e5b43e4f71cef2f79/packages/remirror__extension-yjs/src/yjs-extension.ts#L274)?

whawker (approved):
Minor suggestion above! Otherwise looks good!

ankon:
> Worth adding a warning in the YJS extension when the annotation extension is present, but the disableUndo option is not set?

I think it could be helpful ... but I looked and don't see any "sanctioned" way of logging warnings anywhere else. The only place that touches the console is `RemirrorError`, so I could create one and catch it, but that feels ... dirty. I'm tempted to just leave it with the documentation updates (decided to put a link to the y-prosemirror PR into it), as anyone using yjs and building a collaborative editor will at least skim over that page. 

Another idea could be to leave a "breadcrumb" in the form of a comment somewhere. When I tried figuring out why my annotations got lost I did do quite some single-stepping through the code, but I'm not sure that that matches everyone else's journey.

WDYT?

whawker:
> I think it could be helpful ... but I looked and don't see any "sanctioned" way of logging warnings anywhere else.

Good point, I'll raise this at the next maintainers meeting.

> Another idea could be to leave a "breadcrumb" in the form of a comment somewhere. When I tried figuring out why my annotations got lost I did do quite some single-stepping through the code, but I'm not sure that that matches everyone else's journey.

It's a good place to start I guess, is there a specific line when the annotations get dropped? Like [here](https://github.com/remirror/remirror/blob/main/packages/remirror__extension-yjs/src/yjs-extension.ts#L197) or something? Failing that, a comment in the annotation plugin itself maybe?

ankon:
`formatAnnotations` is only called when there is actual "annotation actions" to be done (explicit redraw + changes). The dropping happens in the filtering here: https://github.com/remirror/remirror/blob/4f59919b9f9058e8cb77ad2fb0d876a452ae2b46/packages/remirror__extension-annotation/src/annotation-plugin.ts#L107

Basically the y-prosemirror undo does its replace (`_typeChanged`, because `UndoManager` touches the YXmlFragment: https://github.com/yjs/y-prosemirror/blob/039be96fb520e972fa37f0e896b585ef19da1beb/src/plugins/sync-plugin.js#L354), and the annotations extension then sees the transaction with a "doc changed" flag, and re-maps all annotations in its cache.

So we could add a note there, like "If you see annotations getting removed here, check the source of the transaction and whether it contains any unexpected 'replace' steps"?

whawker:
> So we could add a note there, like "If you see annotations getting removed here, check the source of the transaction and whether it contains any unexpected 'replace' steps"?

Might want give examples of "unexpected" i.e. replacing the whole doc content

```
If you see annotations getting removed here check the source of the transaction and whether it contains any unexpected 'replace' steps".
For instance: Replace steps that modify the entire document range can cause issues. This can occur when using `undo` provided by the Yjs extension. Consider using the `disableUndo` option of the Yjs extension, if using both Yjs and Annotations
```

whawker (approved):

PR #1567, opened 2022-03-28
chore(extension-react-tables): bump jsx-dom version

strdr4605:
### Description

From https://github.com/proteriax/jsx-dom/blob/master/CHANGELOG.md#700= this update should not affect remirror.

---

When using remirror in frameworks that do not support ESM, it is required to use `jsx-dom-cjs`, but `jsx-dom-cjs` in available only with `jsx-dom@v7+`

### Checklist

- [x] My code follows the code style of this project and `pnpm fix` was completed successfully.
- [x] New code is unit tested and all current tests pass when running `pnpm test extension-react-tables`.

### DIFF
--- a/packages/remirror__extension-react-tables/package.json
+++ b/packages/remirror__extension-react-tables/package.json
@@ -58,7 +58,7 @@
     "@remirror/react-core": "^1.1.0",
     "@remirror/react-hooks": "^1.0.27",
     "@remirror/theme": "^1.2.1",
-    "jsx-dom": "^6.4.23"
+    "jsx-dom": "^7.0.2"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


ocavue:
We didn't update jsx-dom to v7 because jsx-dom v7 is a pure-ESM. Since remirror still needs to support CJS, it cannot depend on pure-ESM dependencies.

> When using remirror in frameworks that do not support ESM, it is required to use `jsx-dom-cjs`. 

I'm not sure how this works. Are you saying remirror should depend on`jsx-dom-cjs@7` instead of `jsx-dom@6` nor `jsx-dom@7`? 

---

Anyway, I'm working on a total rewrite of `extension-react-tables`. We won't need `jsx-dom` at all in the next big version.

PR #1581, opened 2022-04-04
feat(react): support react v18

### LINKED ISSUE #1580: After migration to React 18, EmojiExtension encounters an error.
### Summary
After migration to React 18, EmojiExtension encounters an error.
React 18 works normally when EmojiExtension isn't used.

I'm using 
- @remirror/react 1.0.31
- nextjs 12.1.4
- react 18.0.0

```ts
// ...
import emojiData from 'svgmoji/emoji.json';

// currently in use
const extensions = () => [
  new LinkExtension({
    autoLink: true,
  }),
  new AnnotationExtension(),
  new WhitespaceExtension(),
  new StrikeExtension(),
  new HeadingExtension(),
  new BoldExtension(),
  new ItalicExtension(),
  new ImageExtension({ enableResizing: true }),
  new DropCursorExtension({ color: palette.success }),
  new BlockquoteExtension(),
  new BulletListExtension(),
  new OrderedListExtension(),
  new CodeBlockExtension({
    supportedLanguages: [css, javascript, json, markdown, typescript, java],
    syntaxTheme: 'dracula',
  }),
  new EmojiExtension({ data: emojiData, moji: 'twemoji' }), // here error
  new HorizontalRuleExtension(),
  new CodeExtension(),
  new UnderlineExtension(),
];
```

### Screenshot(s)

image

ocavue:
### Description

1. Updated react and react-dom to v18 as well as some other dependencies.
2. Deprecated `useEffectWithWarning` and remove all the usage from the codebase. This will fix #1580. 
3. Make the third parameter of `useExtension` to be optional when passing a callback function to `useExtension`. This is not a breaking change.
4. Add a new `useExtensionCallback` hook for better TypeScript support. 

TBH, it doesn't feel correct to me that users have to set dependencies in the third parameter.

```ts
// before 
function App() {
  useExtension(
    MyExnteion, 
    ({addHandler}) => addHandler('onChange', onChange),
    [onChange], 
  )
  return ...
}
// or
function App() {
  useExtension(
    MyExnteion, 
    useCallback(({addHandler}) => addHandler('onChange', onChange), []), // <-- this will cause warning from `eslint-plugin-react-hooks`
    [onChange], 
  )
  return ...
}

// after 
function App() {
  useExtension(
    MyExnteion, 
    // you CAN NOT get typing of `addHandler`
    useCallback(({addHandler}) => addHandler('onChange', onChange), [onChange])
  )
  return ...
}
// or
function App() {
  useExtensionCallback(
    MyExnteion, 
    // you CAN get typing of `addHandler`
    useCallback(({addHandler}) => addHandler('onChange', onChange), [onChange])
  )
  return ...
}
```

As the performance concern about the fact that `useExtension` should take a memoized callback function, I think we can do very little thing other than making sure that the docs are clear enough. 

### Checklist


packages/remirror__react-core/src/hooks/use-extension.ts
```diff
@@ -90,8 +88,8 @@ export function useExtension<Type extends AnyExtensionConstructor>(
 ): void;
 export function useExtension<Type extends AnyExtensionConstructor>(
   Constructor: Type,
-  method: UseExtensionCallback<Type>,
-  dependencies: DependencyList,
+  memoizedCallback: UseExtensionCallback<Type>,
+  dependencies?: DependencyList,
```
whawker, line 92: Seems like making this optional, (which it already is line 101) would fix the TypeScript hints for `addHandler`

whawker (commented):
I wonder if useExtensionCallback is a little too generic? I wonder we need something like `useExetensionEvent`?

```ts
function useExtensionEvent<Type extends AnyExtensionConstructor, Key extends keyof GetHandler<OptionsOfConstructor<Type>>> (
  extension: Type,
  event: Key,
  handler: Handler<GetHandler<OptionsOfConstructor<Type>>[Key]>
): void {
  return useExtension(extension, ({ addHandler }) => addHandler(event, handler), [event, handler])
}

function useExample() {
  useExtensionEvent(LinkExtension, 'onClick', useCallback((event, data) => {
    return false;
  }, []))
}
```

package.json
```diff
@@ -215,15 +215,15 @@
...
+    "playwright-testing-library": "^4.2.1",
     "prettier": "^2.3.2",
     "prettier-plugin-packagejson": "^2.2.13",
     "remirror": "1.0.76",
     "rimraf": "^3.0.2",
     "size-limit": "^5.0.5",
     "snapshot-diff": "^0.9.0",
     "testing": "^0.0.4",
-    "typescript": "^4.4.4",
+    "typescript": "~4.5.5",
```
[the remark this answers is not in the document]
ocavue: Could you explain your concern a bit? The bundled Remirror packages on NPM don't depend on typescript so I guess it should have any issues for end users.

packages/remirror__react-core/src/hooks/use-extension.ts
ocavue: Below is the issue I see when using `useExtension` without the third argument:

image

ocavue: Below is my understanding of why `useExtension` with two arguments won't work, which may be incorrect.

`useExtension` has three overloads, shown as below:

```tsx
function useExtension(Constructor); // First overload
function useExtension(Constructor, options: DynamicOptionsOfConstructor); // Second overload
function useExtension(Constructor, callback: UseExtensionCallback, dependencies?); // Third overload
function useExtension(Constructor, optionsOrCallback = undefined, dependencies = []) {
  // actual implementation
}
```

When you pass three arguments to `useExtension`, TS will know that you are using the third overload since it's the only overload that can take three arguments, so that it can know that the type of the second argument is `UseExtensionCallback`. 

However, if you only pass two arguments, then TS won't know if the second argument is `DynamicOptionsOfConstructor` or `UseExtensionCallback`, so it won't be able to give the type of `addHandler`.

packages/remirror__react-core/src/hooks/use-effect-with-warning.ts
```diff
@@ -3,6 +3,8 @@ import { DependencyList, useEffect, useRef } from 'react';
 import warning from 'tiny-warning';
 
 /**
+ * @deprecated
```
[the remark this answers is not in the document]
ocavue: Added

ocavue:
I like the idea of `useExtensionEvent`. I will implement it as well as `useExtensionCustomEvent`.

ocavue:
`useExtensionEvent` is implemented. It looks so clean now! 👍

```diff
- useExtension(EmojiExtension, ({ addHandler }) => addHandler('suggestEmoji', onChange), [onChange]);
+ useExtensionEvent(EmojiExtension, 'suggestEmoji', onChange);
```

packages/remirror__react-core/src/hooks/use-extension.ts
whawker: > Here is the issue I see when using `useExtension` without the third argument:
> 
> image

Ah yes, I wasn't using `useCallback` .

We could fix the above by adding

```ts
export function useExtension<Type extends AnyExtensionConstructor>(
  Constructor: Type,
  method: (callback: UseExtensionCallback<Type>, deps: DependencyList) => UseExtensionCallback<Type>
): void;
```

package.json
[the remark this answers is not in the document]
whawker: Cool, maybe we could add 4.4 check into the pipeline, in a way that ignores the docs?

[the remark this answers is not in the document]
whawker: Not right now I mean, not a blocker for this.

packages/remirror__react-core/src/hooks/use-extension.ts
ocavue: Interesting. Will try it.

package.json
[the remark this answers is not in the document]
ocavue: This is doable but it's not a two minutes task, so I'd like to give it a relatively low priority.

ocavue:
Note to myself:

Replace `useExtension` inside `useMultiPositioner` as below can pass all the tests. But I don't understand why. In fact, I don't understand the functionality of `useMultiPositioner` (yet). 

```
  useExtensionCustomEvent(
    PositionerExtension,
    'positioner',
    useMemo(() => {
      const positioner = getPositioner(positionerRef.current);
      setTimeout(() => {
        setMemoizedPositioner(positioner);
      });
      return positioner;
    }, deps),
  );
```

packages/remirror__react-core/src/hooks/use-extension.ts
ocavue: I fixed it! By switching the second and third overload. I guess this will make TS check the second overload (which is `UseExtensionCallback` now) sooner and everything works fine.

 
image

.changeset/tame-rings-study.md
```diff
@@ -0,0 +1,25 @@
+---
+'@remirror/react-core': minor
+---
+
+Add two new hook `useExtensionEvent` and `useExtensionCustomEvent`. You can use them to add event handlers to your extension. They are simpler and easier to use than the existed `useExtension` hook.
```
whawker, line 5: It seems the only things that are using custom handlers, already have hooks that wrap them. i.e. `useKeymap` `useMultiPositioner` or `useSuggester`.

So I think we should avoid adding `useExtensionCustomEvent` to the public API (keep it for internal uses though). 

That way we avoid having to explain the difference between `handlers` and `customHandlers` in the documentation.

ocavue: OK. I will hide it

ocavue: done

whawker (approved):
Still a little worried about the TypeScript 4.5 thing. But approving.

Excellent work @ocavue!!! 💯 💥

ocavue:
I did some research on the TypeScript 4.5 thing. Most large open-source projects I found accept raising their minimum required TS version without a major version release. However, the minimum TS version would be old enough anyway. 

### DefinitelyTyped
 
> https://github.com/DefinitelyTyped/DefinitelyTyped#support-window
> 
> Definitely Typed only tests packages on versions of TypeScript that are less than 2 years old.
> 
> Currently versions 3.8 and above are tested...

### MUI (Material UI)

> https://mui.com/material-ui/getting-started/supported-platforms/
> MUI requires a minimum version of TypeScript 3.5. This aims to match the policy of [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped), with the support of the versions of TypeScript that are less than two years old.

### Vue.js 

> https://vuejs.org/about/releases.html#release-cycle
>
> We may ship incompatible changes to TypeScript definitions between *minor* versions. This is because:
> 
> Sometimes TypeScript itself ships incompatible changes between minor versions, and we may have to adjust types to support newer versions of TypeScript.
> 
> Occasionally we may need to adopt features that are only available in a newer version of TypeScript, raising the minimum required version of TypeScript.
> 
> If you are using TypeScript, you can use a semver range that locks the current minor and manually upgrade when a new minor version of Vue is released.


> https://github.com/vuejs/core/blob/main/CHANGELOG.md#320-2021-08-09
> 
> ## [3.2.0](https://github.com/vuejs/core/compare/v3.2.0-beta.8...v3.2.0) (2021-08-09)
> Compatibility Notes
> This release contains no public API breakage. However, there are a few compatibility related notes:
> This release ships TypeScript typings that rely on [Template Literal Types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#template-literal-types) and requires TS >= 4.1.

So I feel we could follow their practice. Since we are using TS version 4.4 before this PR, then we could keep our `.d.ts` files compatible with TS 4.4 for a long period of time. When TS 4.4 is old enough and we do want some fancy new features from the new version of TS, we can update our minimal TS requirement, with a minor or major version release. 


---

If we want to promise our minimum required TS version, then it's better to have some kind of CI pipeline to check it. [downlevel-dts](https://www.npmjs.com/package/downlevel-dts) that might be able to fit our requirements. 

--- 

Anyway, this PR itself shouldn't bring TS issues for our users directly, so I will just remove my Changeset about TS version and merge it.

whawker:
Thanks for the TypeScript update notes, very useful!

Issue #1574, opened 2022-03-31
Tab-Key behavior inside the editor changed


VFritsch:
### Summary

In a older version of remirror, pressing the Tab-Key caused the editor to lose focus -> jump to the next element, while pressing it inside a list it added a new indent.

This behavior is not supported anymore in the current version.

Is there any way to enable this behavior again or recreate it manually via code with properties/components?
### Steps to reproduce

Still working in this version:
https://615b820a97b64a25171ada6e--remirror.netlify.app/

### Expected results

There should be a way to replicate the old behavior or allow both events for pressing the tab-key.

### Actual results

Pressing Tab inside the empty editor does nothing, only the identation inside a list is working.

whawker:
Will raise this at the next maintainer meeting. I find this painful too. Related to #1489

whawker:
@VFritsch this should be fixed. Sorry for the delay!

VFritsch:
Thank you very much!

PR #1636, opened 2022-05-06
fix: resolve the shortcut conflict between node formatting and list

### LINKED ISSUE #1489: NodeFormattingExtension stops working when combined with any ListExtension
### Summary

Text indentation implemented in `NodeFormattingExtension` stops working when editor extensions contain one of the list extensions (`BulletListExtension` or`OrderedListExtension`).

### Steps to reproduce

Check out this [codesandbox](https://codesandbox.io/s/cranky-fast-r2ry9?file=/src/App.js). There are two editors there:
- the first one contains only `NodeFormattingExtension` - if you type in anything there and click `Tab` the text will be indented
- the second one contains `NodeFormattingExtension` and `BulletListExtension` - if you type in anything there and click `Tab` the text won't get indented

`@remirror/pm` - `1.0.11`
`@remirror/react` - `1.0.21`
`remirror` - `1.0.61`

### Expected results

Text indentation implemented in `NodeFormattingExtension` works even if there are list extensions (`BulletListExtension` or
  `OrderedListExtension`) added in the same editor.

### Actual results

Text indentation implemented in `NodeFormattingExtension` doesn't work when there is at least one list extension added in the same editor.

ocavue:
### Description

Close #1489 

### Checklist

### Screenshots

 

https://user-images.githubusercontent.com/24715727/169849485-1503dbee-8f89-4bbd-a5f5-a2ae84036e0c.mp4

### DIFF
--- a/packages/remirror__extension-node-formatting/__tests__/node-formatting-extension.spec.ts
+++ b/packages/remirror__extension-node-formatting/__tests__/node-formatting-extension.spec.ts
@@ -1,5 +1,78 @@
-import { extensionValidityTest } from 'jest-remirror';
+import { extensionValidityTest, renderEditor } from 'jest-remirror';
+import { BulletListExtension } from 'remirror/extensions';
 
 import { NodeFormattingExtension } from '../';
 
 extensionValidityTest(NodeFormattingExtension);
+
+describe('keybindings', () => {
+  const extensions = [new NodeFormattingExtension()];
+  const editor = renderEditor(extensions);
+  const {
+    nodes: { doc },
+    attributeNodes: { paragraph: p },
+  } = editor;
+
+  it('can increase and decrease indent', () => {
+    editor.add(doc(p()('hello')));
+
+    editor.press('Tab');
+    expect(editor.state.doc).toEqualRemirrorDocument(doc(p({ nodeIndent: 1 })('hello')));
+    editor.press('Tab');
+    expect(editor.state.doc).toEqualRemirrorDocument(doc(p({ nodeIndent: 2 })('hello')));
+
+    editor.press('Shift-Tab');
+    expect(editor.state.doc).toEqualRemirrorDocument(doc(p({ nodeIndent: 1 })('hello')));
+    editor.press('Shift-Tab');
+    expect(editor.state.doc).toEqualRemirrorDocument(doc(p({ nodeIndent: 0 })('hello')));
+  });
+});
+
+describe('keybindings with list extensions', () => {
+  const extensions = [new NodeFormattingExtension(), new BulletListExtension()];
+  const editor = renderEditor(extensions);
+  const {
+    nodes: { doc, listItem: li, bulletList: ul },
+    attributeNodes: { paragraph: p },
+  } = editor;
+
+  it('ignore the indentation keybindings in list items', () => {
+    const doc1 = doc(
+      ul(
+        //
+        li(p()('A')),
+        li(p()('B')),
+        li(p({ nodeIndent: 0 })('C<cursor>')),
+        li(p()('D')),
+      ),
+    );
+
+    const doc2 = doc(
+      ul(
+        //
+        li(p()('A')),
+        li(
+          //
+          p()('B'),
+          ul(
+            //
+            li(p({ nodeIndent: 0 })('C<cursor>')),
+          ),
+        ),
+        li(p()('D')),
+      ),
+    );
+
+    editor.add(doc1);
+
+    editor.press('Tab');
+    expect(editor.state.doc).toEqualRemirrorDocument(doc2);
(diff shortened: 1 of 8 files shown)


packages/remirror__extension-node-formatting/__tests__/node-formatting-extension.spec.ts
```diff
@@ -1,5 +1,96 @@
-import { extensionValidityTest } from 'jest-remirror';
+import { extensionValidityTest, renderEditor } from 'jest-remirror';
+import { BulletListExtension } from '@remirror/extension-list';
```
whawker, line 2: Can we

```suggestion
import { BulletListExtension } from 'remirror/extensions';
```

To avoid the need for the new dependency in `package.json`?

packages/remirror__extension-node-formatting/src/node-formatting-extension.ts
```diff
@@ -207,12 +208,20 @@ export class NodeFormattingExtension extends PlainExtension<NodeFormattingOption
...
+  })
   increaseIndentShortcut(props: KeyBindingProps): boolean {
     return this.increaseIndent()(props);
   }
 
-  @keyBinding({ shortcut: NamedShortcut.DecreaseIndent, command: 'decreaseIndent' })
+  @keyBinding({
+    shortcut: NamedShortcut.DecreaseIndent,
+    command: 'decreaseIndent',
+    priority: ExtensionPriority.High,
```
whawker, line 223: Why is this one High is the other is Low priority?

ocavue: - When pressing tab
  - Try to indent list item first (ExtensionPriority.Default)
  - If not possible, then indent node formatting (ExtensionPriority.Low)
- When pressing shift-tab
  - Try to dedent node formatting first (ExtensionPriority.High)
  - If not possible, then dedent list item (ExtensionPriority.Default)

ocavue: My idea is to make `tab` and `shift-tab` symmetric. 

For a pair of shortcuts which do opposite operations, by pressing one shortcut, you kind of add a new state into a stack, and pressing the other shortcut will pop the lastest state for the stack and make you have the previous state. 

We have many pairs of shortcuts that match this rule, for example Undo - Redo, Zoom In - Zoom Out, Left Arrow - Right Arrow. 

```
       --- KeyA -->        --- KeyA -->        --- KeyA --> 
state1              state2              state3              state4
       <-- KeyB ---        <-- KeyB ---        <-- KeyB ---
```

https://user-images.githubusercontent.com/24715727/167158142-f3bcd3d9-6aaf-4674-aca5-7c6e54b6f90e.mp4

whawker: That wouldn't be true though right?

If I was within a list, in an indented parent 

```
blockquote(indent=20)
  bulletList
    listitem
      paragraph
    listItem
      paragraph<cursor>
```

and press tab, I indent list item. 

If I then press shift-tab I end up dedenting the entire parent blockquote? Seems wrong to me?

whawker: Unless all the commands check only the current depth?

whawker: @ocavue I think both our concerns could be addressed by having

When pressing shift-tab
* Try to dedent node formatting at **current depth only** (ExtensionPriority.Medium)
* Try to dedent list item (ExtensionPriority.Default)
* If not possible, then dedent node formatting of any parent (ExtensionPriority.Low)

Thoughts?

ocavue:
@whawker I pushed another approach. We by default ignore all tab / shift-tab keybindings from node-formatting if the cursor is in a list.

packages/remirror__extension-node-formatting/src/node-formatting-extension.ts
```diff
@@ -51,7 +53,7 @@ import {
       '180px',
       '200px',
     ],
-    excludeNodes: [],
+    excludeNodes: ['bulletList', 'listItem'],
```
whawker, line 56: do we need to add `orderedList`, `taskList` and `taskListItem` here?

It is possible to use the group name here? i.e. `ListContainerNode` and `ListItemNode`?

ocavue: Added `orderedList`, `taskList` and `taskListItem`.

We have to edit `@remirorr/core` to support group name here. I kind of don't want to touch `core` unless very necessary.

whawker:
Fixes #1489. Fixes #1574

PR #1648, opened 2022-05-17
feat: support highlighting of content

ronnyroeller:
### Description

We've been using the annotation-extension to highlight content. Yet, this has proven to create lots of issues in combination with Collaborative Editing (see details in this [blog post](https://medium.com/collaborne-engineering/prosemirror-highlights-comments-20ce820149ed)).

This PR contributes our current highlights-as-marks extension. The draft PR still misses storybooks, documentations, etc. - and is mainly intended to spark a discussion how to contribute the extension.

### Relationship to `TextHighlightExtension`

I'd like to discuss especially how this relates to the existing [TextHighlightExtension](https://remirror.io/docs/extensions/text-highlight-extension), which provides a simpler version of what we need. Notable, text-highlight adds simply a background color to a piece of text. In contrast, the extension in this PR adds an ID to the highlight. This allows to add meta data that can be managed somewhere else (e.g. a comments on the highlight, tags, etc). As a side effect, this means that there can be multiple, different highlights with the same color. It also allows for highlights (called "joined highlights") with multiple text parts (called "disjoined highlights" in the code). Additionally, the PR allows for overlapping highlights - while text-highlight today allows only for a single highlight at a position.

So, one of my key questions is if we should try to make `text-highlight` more generic, so it covers both use cases (color only AND meta data), or if we should rather have two different extensions. I tend more towards extending `text-highlight` to avoid confusions (like we have with `table-extension` and `react-table-extension`). But I'm not sure if it's technical possible without breaking backwards-compatibility...

### Implementation proposal

We could extend the `TextHighlightExtension` as following:
- Users would define the list of attributes, e.g. `highlight`, `id`, `tags`, etc. => This would require also custom toDom, parseDom for these attributes.
- Users could pass these attributes when creating/updating highlights -> Q: Can we ensure type safety?
- The queryHighlights() function gets a function to detect if different disjoint highlights belong together, e.g. `isSame(highlight1, highlight2) => highlight1.id === highlight2`.

What I'm doubting about: Does this create so much overhead (e.g. for attribute-specific toDom/parseDom) that two separate extensions would be simpler overall...?

### Checklist


packages/remirror__extension-highlight/src/utils/joined-highlights.ts
```diff
@@ -0,0 +1,51 @@
...
+  disjoinedHighlights: HighlightAttrs[],
+): HighlightAttrs[] => {
+  const joinedHighlights: HighlightAttrs[] = [];
+
+  for (const disjoinedHighlight of disjoinedHighlights) {
+    joinDisjoinedHighlight(joinedHighlights, disjoinedHighlight);
+  }
+
+  return joinedHighlights;
+};
```
whawker, line 51: This seem very convoluted, you could achieve the same by looking through node and interrogating each text node.

[see here](https://codesandbox.io/s/intelligent-stonebraker-00vgkt?file=/src/HighlightExtension.ts:1738-2785)

packages/remirror__extension-highlight/src/highlight-extension.ts
```diff
@@ -0,0 +1,304 @@
...
+  },
+})
+export class HighlightExtension extends MarkExtension<HighlightOptions> {
+  get name(): string {
+    return 'highlight' as const;
+  }
+
+  createMarkSpec(extra: ApplySchemaAttributes, override: MarkSpecOverride): MarkExtensionSpec {
+    const idAttr = `id`;
+    const tagsAttr = `tags`;
```
whawker, line 59:
```suggestion
    const tagsAttr = `data-tags`;
```

For valid HTML?

whawker (commented):
### Naming

I think this extension should be an "EntityReferenceExtension", as we are linking some external entity to the document text. The important part here is the `id`, I could see this being used to implement "comments" for instance.

### Attributes

I don't believe `tags` should be part of this extension, one because it i's very specific for Next's use case, and two I don't think it's actually necessary?

I would propose that we could pass an "entity by ID" lookup function as an extension option to enrich, or alternatively this lookup could be performed in the _positioner_ instead.

I think the previous issues with _annotations_ were the `to` and `from` position being separate from the document structure and difficult to keep in sync, not so much the highlight metadata itself? 

Rendering them as marks solves the `from` and `to` issue, but adding tags as well seems like a step too far? Perhaps I am oversimplifying for the collaboration use case though?

### Extracting highlights

I think the current implementation is convoluted - I think [this approach](https://codesandbox.io/s/intelligent-stonebraker-00vgkt?file=/src/HighlightExtension.ts:1738-2785) is simpler and easier to maintain?

I think the CodeSandbox approach could be easily extended for the "getHighlightsAtPos" functionality by using the other two arguments available in [`getDescendents`](https://prosemirror.net/docs/ref/#model.Node.descendants)

packages/remirror__extension-highlight/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@remirror/extension-highlight",
+  "version": "1.0.1",
+  "private": true,
```
whawker, line 4: Remember to remove this before merging

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -0,0 +1,264 @@
...
+
+    return {
+      ...override,
+      excludes: '',
+      attrs: {
+        ...extra.defaults(),
+        id: { default: '' },
+      },
+      // Note that we don't implement `ParseDom`.
+      // The reasoning behind this is to disable pasting (parsing) the entityReference marks in the editor.
```
whawker, line 82: I think we should `parseDOM` by default - for consistency with every other extension, but give the user to opt out.

```ts
const spec = {
  // stuff here
};
if (!this.option.disableParseDOM) { // double negative - maybe confusing?
  spec.parseDOM = [
    {
      tag: `span[${idAttr}]`,
      getAttrs: (node) => {
        if (!isElementDomNode(node)) {
          return false;
        }

        const id = node.getAttribute(idAttr);
        return { ...extra.parse(node), id };
      },
    },
    ...(override.parseDOM ?? []),
  ]
}
return spec
```

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -0,0 +1,264 @@
...
+        },
+      },
+    };
+  }
+
+  @command()
+  addEntityReference(id?: string): CommandFunction {
+    return ({ state, tr, dispatch }) => {
+      const { from, to } = state.selection;
+      const newId = id ?? (this.options.createId ? this.options.createId() : uniqueId());
```
whawker, line 120: `createId` already defaults to `uniqueId`

```suggestion
      const newId = id ?? this.options.createId();
```

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -0,0 +1,264 @@
...
+    doc.descendants((node: Node, pos: number) => {
+      const subDisjoinedEntityReferences = getDisjoinedEntityReferencesFromNode(
+        node,
+        pos,
+        this.name,
+      );
+
+      if (subDisjoinedEntityReferences.length === 0) {
+        return true;
+      }
```
whawker, line 187: I suspect this is an over simplification. I suspect we should 

```ts
if (node.isAtom) {
  return false;
}

if (!node.isText) { // Assuming the only thing that can have marks is text nodes? I might be wrong here
  return true;
}

const subDisjoinedEntityReferences = getDisjoinedEntityReferencesFromNode(
  node,
  pos,
  this.name,
);
```

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -0,0 +1,264 @@
...
+   * entityReferences. Therefore, a entityReference that spans multiple paragraphs is
+   * internally stored as multiple marks (all having the same entityReference ID as
+   * attribute).
+   *
+   * To get the actual entityReferences (for which each entityReference ID is unique), call
+   * `joinDisjoinedEntityReferences`.
+   */
+  @helper()
+  getEntityReferences(): Helper<EntityReferenceMetaData[]> {
+    const entityReferences = getEntityReferencesFromPluginState({
```
whawker, line 208: Could just use `this.getPluginState()` here?

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -0,0 +1,264 @@
...
+        },
+        0,
+      ],
+    };
+  }
+
+  /**
+   * Create the extension plugin for inserting decorations into the editor.
+   */
+  createPlugin(): CreateExtensionPlugin {
```
whawker, line 96: We should use the generic type here


```ts
interface EntityReferenceState {
  entityReferences: { [key: string]: any };
}

createPlugin(): CreateExtensionPlugin<EntityReferenceState> {
```

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -0,0 +1,264 @@
...
+   *
+   * @returns all entityReferences at a specific position in the editor.
+   *
+   * In order to use this helper make sure you have the
+   * [[`EntityReferenceExtension`]] added to your editor.
+   */
+  @helper()
+  getEntityReferencesAt(
+    pos?: PrimitiveSelection,
+    includeEdges = true,
```
whawker, line 227: I think we should drop this option still

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -0,0 +1,264 @@
...
+      return true;
+    };
+  }
+
+  @command()
+  redrawEntityReferences(): CommandFunction {
+    return ({ tr, dispatch }) => {
+      dispatch?.(
+        tr.setMeta(EntityReferenceExtension.name, {
+          type: ActionType.REDRAW_ENTITY_REFERENCES,
```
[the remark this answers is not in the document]
IdrissMahjoubi: The reason behind this command is to force update the editor (e.i recalculate decorations) outside of the editor. Is there a better way to do this?

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
IdrissMahjoubi: There are some use cases where we might need to reference a non-text node.

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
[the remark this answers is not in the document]
whawker: Ah apologies @IdrissMahjoubi, I was expecting to see something in the plugin listening to this meta type? It may be possible to use the existing [`forceUpdate`](https://github.com/remirror/remirror/blob/main/packages/remirror__core/src/builtins/commands-extension.ts#L520-L526) command to fulfil this use case?

---

P.s. are decorations necessary anymore? As marks already get rendered into the document?

I wonder if `addEntityReference` should also allow for a `color` mark attribute? To determine how it is rendered in document?

P.s. If marks can be used in this way, it might be worth looking at the CSS property [`mix-blend-mode`](https://developer.mozilla.org/en-US/docs/Web/CSS/mix-blend-mode) to determine the color behaviour when there are overlapping entity references

cc @ronnyroeller for context on color overlapping behaviour.

packages/remirror__extension-entity-reference/src/utils/joined-entity-references.ts
```diff
@@ -0,0 +1,57 @@
...
+  // Find outer bound of all marks belong to the entityReference
+  const [same, diff] = partitionEntityReferences(entityReferences, (h) => h.id === id);
+  const [from, to] = findMinMaxRange([...same, entityReference]);
+  let fullText = same.map((h) => h.text).join(' ');
+  // Respect existing keys and merge them into the new entityReference.
+  const newEntityReference: EntityReferenceMetaData = {
+    ...entityReference,
+    from,
+    to,
+    text: (fullText += text),
```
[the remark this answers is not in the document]
ronnyroeller: We should keep it like this for now. It's mathematically correct to just combine them. Maybe we see different use cases in the future, and can create a follow up PR to change this behavior.

What we should still is to change to `text: fullText.concat(text)` and then make `fullText` a `const`

packages/remirror__extension-entity-reference/__tests__/entity-reference-extension.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  it('Can overlap partially with other entityReferences', () => {
+    const DUMMY_TEXT = 'Lorem Ipsum is simply dummy text of the printing and typesetting industry.';
+    const editor = add(doc(p(DUMMY_TEXT)));
+    editor.selectText({ from: 1, to: 38 }).commands.addEntityReference();
+    editor.selectText({ from: 35, to: 74 }).commands.addEntityReference();
+    const createdEntityReferences = editor.helpers.getEntityReferences();
+    expect(createdEntityReferences).toHaveLength(2);
+    const createdEntityReferenceIds = createdEntityReferences.map((h) => h.id);
+    expect(isArrayUnique(createdEntityReferenceIds)).toBe(true);
```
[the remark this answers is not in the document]
ronnyroeller: Cool. Will fix

packages/remirror__extension-entity-reference/__tests__/entity-reference-extension.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+  } = create();
+
+  it('Can overlap partially with other entityReferences', () => {
+    const DUMMY_TEXT = 'Lorem Ipsum is simply dummy text of the printing and typesetting industry.';
+    const editor = add(doc(p(DUMMY_TEXT)));
+    editor.selectText({ from: 1, to: 38 }).commands.addEntityReference();
+    editor.selectText({ from: 35, to: 74 }).commands.addEntityReference();
+    const createdEntityReferences = editor.helpers.getEntityReferences();
+    expect(createdEntityReferences).toHaveLength(2);
+    const createdEntityReferenceIds = createdEntityReferences.map((h) => h.id);
```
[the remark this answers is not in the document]
ronnyroeller: Will fix

packages/remirror__extension-entity-reference/__tests__/entity-reference-extension.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+ * @returns True if the array contains only unique values
+ */
+const isArrayUnique = (arr: string | Iterable<any> | null | undefined) =>
+  Array.isArray(arr) && new Set(arr).size === arr.length;
+
+describe('EntityReference marks', () => {
+  const {
+    add,
+    nodes: { p, doc },
+  } = create();
```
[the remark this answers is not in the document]
ronnyroeller: Good one. We actually don't even need it for testing. The ID can now be passed directly to `addEntityReference`, so the caller can control the ID. This is sufficient for testing.
We'll remove it completely to keep the Options clean.

packages/remirror__extension-entity-reference/__tests__/entity-reference-extension.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+  return renderEditor([new EntityReferenceExtension()]);
+}
+
+/**
+ * @returns True if the array contains only unique values
+ */
+const isArrayUnique = (arr: string | Iterable<any> | null | undefined) =>
+  Array.isArray(arr) && new Set(arr).size === arr.length;
+
+describe('EntityReference marks', () => {
```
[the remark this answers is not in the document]
ronnyroeller: Good point. We'll do this similar to the other extension tests (e.g. https://github.com/remirror/remirror/blob/c4955566b83553a7b760886c8f1d1c2105ba487f/packages/remirror__extension-bold/__tests__/bold-extension.spec.ts#L17)

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -0,0 +1,264 @@
...
+    createId: uniqueId,
+  },
+})
+export class EntityReferenceExtension extends MarkExtension<EntityReferenceOptions> {
+  get name(): string {
+    return 'entity-reference' as const;
+  }
+
+  createMarkSpec(extra: ApplySchemaAttributes, override: MarkSpecOverride): MarkExtensionSpec {
+    const idAttr = `data-entityReference`;
```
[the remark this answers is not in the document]
ronnyroeller: Will fix

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -0,0 +1,264 @@
...
+        ...extra.defaults(),
+        id: { default: '' },
+      },
+      // Note that we don't implement `ParseDom`.
+      // The reasoning behind this is to disable pasting (parsing) the entityReference marks in the editor.
+      toDOM: (mark: Mark) => [
+        'span',
+        {
+          [idAttr]: mark.attrs.id,
+        },
```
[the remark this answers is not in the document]
ronnyroeller: Will fix

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
ronnyroeller: Indeed, this will be obsolete when we remove `createId`

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
ronnyroeller: A better is actually to always have the `parseDom` in the `EntityReferenceExtension`.
For our app, we can just inherit the class, and overwrite `createMarkSpec` accordingly.
This way, the disabling doesn't leak into the Remirror extension itself.

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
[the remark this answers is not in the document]
ronnyroeller: TODO:
- Check if `forceUpdate` can do the same
- If not: Add a comment to the command to explain what it does (force redraw if an external color for a entity is changed) and how it works (inject a change to the document) and why this is required to be implemented like that

> I wonder if addEntityReference should also allow for a color mark attribute?

This might work for simple use case but wouldn't work for what we need:
- Our users can change the color via the entity. -> This would require us to change the color in all mark attributes across all Prosemirror documents, whenever the color of an entity changes.
- Our entities have semantic colors, which are then translated to hex colors based on light/dark mode. -> This would require us to change the color in all mark attributes across all Prosemirror documents, whenever the user switches between light/dark mode.

So, therefore I'd keep `color` out of scope for now (to keep the extension as simple as possible). Maybe somebody could add in the future, if there is a real use for it.

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
ronnyroeller: Good point. We don't need the config option ourselves, so will remove it for now to keep the extension simple.

packages/remirror__extension-highlight/src/utils/joined-highlights.ts
IdrissMahjoubi: Thank you for the insight. It's true that it looks convoluted. The idea behind getting the disjoined marks is to create decorations based on them (e.g blending the colors of the overlapping marks). overlapping marks, in this case, are represented as a subArray in the return value of `getDisjoinedEntityReferences` helper. I guess we could somehow "disjoin" them using the overlapping ranges after we get the _joined marks_ (using `extractHighlightMarks` from the sandbox). Do you think that's an optimal way to do it?

ronnyroeller:
Thanks again for the review, @whawker! All feedback has been processed. Please let me know if you see any other points.

packages/remirror__extension-entity-reference/__tests__/entity-reference-extension.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+    commands.addEntityReference();
+
+    const createdEntityReferences = helpers.getEntityReferences();
+    expect(createdEntityReferences).toHaveLength(2);
+
+    expect(createdEntityReferences).toEqual(uniqueArray(createdEntityReferences));
+  });
+
+  it('Can overlap completely with other entityReferences', () => {
+    const DUMMY_TEXT = 'Lorem Ipsum is simply dummy text of the printing and typesetting industry.';
```
whawker, line 84: Minor: duplicate

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -0,0 +1,247 @@
...
+    };
+  }
+
+  /**
+   * Create the extension plugin for inserting decorations into the editor.
+   */
+  createPlugin(): CreateExtensionPlugin<EntityReferenceState> {
+    return {
+      state: {
+        init: (_: { [key: string]: any }, state: EditorState) => {
```
whawker, line 111:
```suggestion
        init: (_: EntityReferenceState, state: EditorState) => {
```

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -0,0 +1,247 @@
...
+          const entityReferences = this.getDisjoinedEntityReferences(state.doc);
+          return { entityReferences };
+        },
+        apply: (_tr: Transaction, _value: any, _oldState: EditorState, newState: EditorState) => {
```
whawker, line 115:
```suggestion
        apply: (_tr: Transaction, _value: EntityReferenceState, _oldState: EditorState, newState: EditorState) => {
```

packages/remirror__extension-entity-reference/src/types.ts
```diff
@@ -0,0 +1,42 @@
+import { Decoration } from 'remirror';
+import type { AcceptUndefined } from '@remirror/core';
+
+export interface Range {
```
whawker, line 4: This is already a type in core `FromToProps`

whawker (approved):
Some minor last things. But otherwise approved!

packages/storybook-react/stories/extension-entity-reference/highlights.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const highlightsAt = getEntityReferencesAt();
+  return (
+    <>
+      {ALL_HIGHLIGHT_TYPES.map((type) => {
+        const highlightsOfType = highlightsAt.filter((h) => {
+          const highlightType = allHighlights.get(h.id);
+          return highlightType === type;
+        });
+        // Provide visual feedback if there is a highlight of this type at the user's cursor
+        const active = highlightsOfType.length > 0;
```
whawker, line 55: Just curious about this check, it means you can't nested entity references. Is feel like it's misleading.

Assuming text formatting here as entity references (sorry best I can do in markdown)

_A paragraph of text with a **nested entity reference** within it_

packages/storybook-react/stories/extension-entity-reference/highlights.tsx
```diff
@@ -0,0 +1,106 @@
...
+      })}
+    </>
+  );
+};
+
+const Highlights = (): JSX.Element => {
+  const { manager, state, onChange } = useRemirror({
+    extensions: extensions,
+    content: '<p>Highlight important and interesting text</p>',
+    stringHandler: htmlToProsemirrorNode,
```
whawker, line 88:
```suggestion
    stringHandler: 'html',
```

packages/storybook-react/stories/extension-entity-reference/highlights.tsx
ronnyroeller: You can. For example the storybook (cursor is on "and"):
image

whawker: But you can't use the "important" tag twice. The second time removes the first one.

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
ronnyroeller: Hmm, Typescript doesn't like this:
image

Do you know why?

packages/storybook-react/stories/extension-entity-reference/highlights.tsx
ronnyroeller: Aha, now I get it. Indeed, nesting isn't possible in the example. I didn't really know how to keep the example minimal otherwise (which would require multiple separate remove buttons).
You think the demo still brings the point across? Shall we just add a comment to point out this behavior? Or you think the storybook should be different (e.g. have remove buttons instead of just toggle button)?

whawker: A remove button could work. I was also wondering if we could compare the start and end positions of an existing reference too?

I.e. if the positions are same, delete it

whawker:
Also been thinking, we should probably expose a `clickMark` handler, so you could load the actual referenced entity when you click the mark

ronnyroeller:
> Also been thinking, we should probably expose a `clickMark` handler, so you could load the actual referenced entity when you click the mark.

We found in our own usage positioners to be the better approach:
1. Clicking the mark will lead Remirror to set the cursor within the mark (standard behavior)
2. The positioner can use `getEntityReferencesAt` to find all referenced entities and pick one (e.g. the shortest mark).

This approach is demonstrated by the toggle buttons in the storybook.

We found two drawbacks of click handlers:
- It limits keyboard support. With positioners, the user can use left/right keys to trigger the details of the entity. Click handlers require the user to do this via a touch input.
- Positioners feel more natural as that's a more common way of doing things in Remirror than clickhandlers.

Above are just the learnings from our own usage. Happy to add click handler if you see a use case for which positioners wouldn't be a good solution.

packages/storybook-react/stories/extension-entity-reference/highlights.tsx
ronnyroeller: I thought about this a bit further, and I increasingly think that we shouldn't change the storybook because:

**1. For highlights (ie. the storybook example)**
We know from our own usage over the last years that this a theoretical option. We haven't seen any user highlight a sentence as "important" and then as well a word within the same sentence as "important". Users might highlight nested content with different entities/tags - but with the same is quite a corner case. 
Making removing a separate actions would make the storybook considerable more complex because we have to handle than also "which highlight of this type should be removed", etc.
The main purpose of the storybooks is to demonstrate common usage in the simplest possible way (e.g. without fancy styling, error handling, etc). My feeling is that we would serve this here better by not covering the corner case.

**2. For non-highlights:**
I'm not sure if referencing the same entity from nested content is relevant or not. We haven't see it but I would also not rule it out. So, I'd keep it as a possibility in the extension (aside that forbidding would actually make the extension more complex) but wait with adding a storybook for that case until we have a real-world use case.

Does this reasoning make sense to you?

> I was also wondering if we could compare the start and end positions of an existing reference too?

Comparing start/end position is tricky because references might have "holes". See below an example from Google Docs where the yellow comment is split in two parts:
image

These holes can happen if you copy & paste parts of the text with a mark. One could do funky things like also copying the entities (e.g. one comment become two comments) but this isn't always what the user expects ("Why do I have now two comments?") and it adds a lot of follow up complexity (e.g. undo/redo required to split/re-join entities).

whawker: Totally get your point on number 1 and agree we should keep it simple.

For non highlight entities, (specifically comments) I imagined a paragraph of text with bad grammar, and a single misspelled word.

I imagine a user selecting the entire paragraph and commenting. 
Then, selecting the misspelled word (nested in the paragraph) and adding another comment on that. 

In this use case they would have different entity IDs so maybe, this is already covered?

whawker:
Really interesting findings on positioners, I hadn't considered the arrow keys use case.

Could we export the positioner logic like we do with annotations

https://github.com/remirror/remirror/blob/a2ca7a83f35b3831b97817eb2cb38b1a82d60ab8/packages/remirror__extension-annotation/src/annotation-positioners.ts#L26

packages/storybook-react/stories/extension-entity-reference/highlights.tsx
ronnyroeller: Yes, indeed. Different comments would be different entities.

ronnyroeller:
> Could we export the positioner logic like we do with annotations

Added via https://github.com/remirror/remirror/pull/1648/commits/5e8f56eeb69df2b0638bc2ed35f179d76e782e89

ronnyroeller:
@whawker - I believe we addressed now all raised question. Please let me know if you see anything else, or if we are ready to merge? (I know you had already earlier approved the PR but want to make sure to not miss anything.)

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
whawker: @ronnyroeller I just wanted to have a quick look into this before we merge. Happy if I push to this branch?

ronnyroeller: Sure, go ahead

whawker:
@ronnyroeller changes added. I'm happy for this to be merged

ronnyroeller:
> @ronnyroeller changes added. I'm happy for this to be merged

Thanks so much for working through this together, @whawker! 🙏

PR #1654, opened 2022-05-23
feat(react,core): simplify how to subscribe to document changes

whawker:
### Description

Add a hook, and 2 React components to simplify subscribing to document changes. This is simplify our "getting started" for a very common use case. (Docs to be updated if others are happy with this approach)

Adds a `useDocChanged` hook, which when given a handler function, calls it with the transactions and state when document is changed.

```js
import { useCallback } from 'react';
import { useDocChanged } from '@remirror/react';

useDocChanged(
  useCallback(({ tr, transactions, state }) => {
    console.log('Transaction', tr);
    console.log('Transactions', transactions);
    console.log('EditorState', state);
  }, []),
);
```

Also adds two React components, `OnChangeJSON` and `OnChangeHTML` which accept a handler function, which is called with the JSON or HTML serialisation of doc state, whenever the document is changed.

```jsx
import { useCallback } from 'react';
import { OnChangeJSON, OnChangeHTML } from '@remirror/react';

const handleChangeJSON = useCallback((json) => {
  console.log('JSON serialised state', json);
}, []);

const handleChangeHTML = useCallback((html) => {
  console.log('HTML serialised state', html);
}, []);

return (
  <Remirror manager={manager} autoRender>
    <OnChangeJSON onChange={handleChangeJSON} />
    <OnChangeHTML onChange={handleChangeHTML} />
  </Remirror>
);
```

### Checklist

- [ ]  I have updated the documentation where necessary.


ocavue (approved):
LGTM

ocavue:
I'm thinking maybe we should use something like [`use-event-callback`](https://www.npmjs.com/package/use-event-callback). By doing that, we can get rid of `useCallback` and also reduce those unnecessary re-renders caused by some dependencies changing from `useCallback`.

See also: [React RFC for useEvent](https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md)

whawker:
@ocavue Interesting. Shall we explore that in issue separate to this though? 

As an aside, should we deprecate Remirror `useEvent` in favour of some new name (same functionality, new name) - to avoid clashing with the (React) `useEvent` RFC

ocavue:
> Shall we explore that in issue separate to this though?

Agree. It's a separate topic and is not dirrecly related to this PR. 

> should we deprecate Remirror useEvent in favour of some new name.

Sounds good. Maybe `useEditorEvent`.

PR #1673, opened 2022-05-31
chore: rename `useEvent` to `useEditorEvent`, standardize parameters

whawker:
### Description

* Rename `useEvent` to `useEditorEvent` to avoid name clash and possible confusion, with the upcoming React hook.
* Remove deprecated `useEvents`
* Standardize the `contextmenu` and `hover` events to return the event (only) as the first parameter, and meta object as the second parameter.
* 
### Checklist


.changeset/large-socks-live.md
```diff
@@ -0,0 +1,8 @@
+---
+'@remirror/extension-react-tables': patch
+'@remirror/react-hooks': patch
```
whawker, line 3: @ocavue should these be major? I wanted to avoid accidentally bumping to 3.0

ocavue: I think this should be major. Changeset will submit a pull request, which includes the update for the `version` field in `package.json`. So we should be able to review the version number and avoid unexpected 3.0 before the package is published to NPM.

PR #1705, opened 2022-06-06
fix(react-tables): fix initialisation on tables on view, and copy and paste bugs

whawker:
### Description

Initialising editor content from HTML or Markdown, that included a React Table, did not initialise the table properly (tables controllers were never injected)

Additionally, copy and pasting a React table didn't check to see if controlled cells were already present. Leading them to be duplicated. (cc @ronnyroeller) 

### Checklist

### DIFF
--- a/.changeset/three-scissors-deny.md
+++ b/.changeset/three-scissors-deny.md
@@ -1,0 +1,13 @@
+---
+'@remirror/extension-markdown': minor
+'@remirror/extension-react-tables': minor
+'@remirror/extension-tables': patch
+'@remirror/styles': patch
+'@remirror/theme': patch
+---
+
+Add a customisible floating button to completely delete React tables.
+
+Fix creating React tables from markdown initial state.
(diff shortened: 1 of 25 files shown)


packages/remirror__pm/package.json
```diff
@@ -185,6 +185,18 @@
   "publishConfig": {
     "access": "public"
   },
+  "@remirror": {
+    "tsconfigs": {
+      "src": {
+        "compilerOptions": {
+          "baseUrl": ".",
+          "paths": {
+            "prosemirror-tables": "node_modules/prosemirror-tables/src"
```
ocavue, line 194: Why do we need this? Is this related to the compilation error we saw before?

whawker: That was the theory. But looks like it doesn't work.

I was trying to avoid using the `prosemirror-tables/dist` which contains the error we're seeing

PR #1713, opened 2022-06-10
chore: migrate to ESM

ocavue:
### Checklist


whawker:
@ocavue where do we stand on commands like `pnpm fix` - it looks like it runs `preconstruct fix` but not `remirror-cli build`

Similarly `pnpm dev` runs `preconstruct watch` but not `remirror-cli watch`

Should the fix/dev commands be updated or removed?

ocavue:
I need to clean these commands as well as update `contributing.md`. It's still a bit mess here.

PR #1734, opened 2022-06-21
feat: autolink trigger character

DiegoPayan:
### Description

Issue: There are some cases where link marks are created by `autolink` in undesired situations, for example in `window.confirm` => `window.co` is a link followed by the text `nfirm`, this is specially annoying while writing code. 

Proposed solution: One way to fix this is to have a trigger character or a set of them and only create the mark if it's at the end of the string. In order to achieve this, I added a new optional property to `LinkExtension` that receives a regex, it'll be placed at the end of `autoLinkRegex` and the implementation will depend on each user/team needs. You can either define a single trigger character or a combination of them.

A bit more of context: It was possible to do this by passing a custom `autoLinkRegex` but we lost that after the TLD validations since `.com ` is not a valid TLD.

### Checklist

### Screenshots

Without `autoLinkAfter`:

ezgif com-gif-maker

With `autoLinkAfter`:

ezgif com-gif-maker (2)

### DIFF
--- a/.changeset/hungry-penguins-move.md
+++ b/.changeset/hungry-penguins-move.md
@@ -1,0 +1,6 @@
+---
+'@remirror/extension-link': patch
+'storybook-react': patch
+---
+
+Add new logic to allow link mark creation only after a trigger character or a combination was found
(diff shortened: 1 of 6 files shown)


whawker:
Thanks for the contribution @DiegoPayan, sorry for the delay reviewing this, I'm currently on holiday.

Will pick this up as soon as I can

DiegoPayan:
I'm closing this PR since #1817 is already approved, and it solves the same problem!

PR #1741, opened 2022-06-23
fix(link): treat links with same href and text content as auto link

### LINKED ISSUE #1740: Add the ability to edit link marks when the text is equal to the URL
### Description

Having the ability to edit each item in the editor before sending a message is important for users. Unfortunately, there are some use cases where there is more than one editor in different contexts, for example in a chat application, if you have a list of users, you want to have the ability to type something, switch to another chat and come back to continue typing in the first one. The problem is that when we restore information in the editor, links are considered as manually created, which makes it not possible to edit them.

marcoSven:
### Description

When href equals text content there is no reason why a link
should not be editable.

closes #1740

### Checklist

### DIFF
--- a/packages/remirror__extension-link/src/link-extension.ts
+++ b/packages/remirror__extension-link/src/link-extension.ts
@@ -296,9 +296,10 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
             // e.g [test](//test.com) - not "auto link"
             // e.g [test.com](//test.com) - "auto link"
             const auto =
-              node.hasAttribute(AUTO_ATTRIBUTE) ||
-              href === text ||
-              href?.replace(`${this.options.defaultProtocol}//`, '') === text;
+              this.options.autoLink &&
+              (node.hasAttribute(AUTO_ATTRIBUTE) ||
+                href === text ||
+                href?.replace(`${this.options.defaultProtocol}//`, '') === text);
 
             return {
               ...extra.parse(node),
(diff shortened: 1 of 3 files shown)


packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -290,7 +290,16 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
-            const auto = node.hasAttribute(AUTO_ATTRIBUTE);
+            const text = node.textContent;
+
+            // If link text content equals href value we "auto link"
+            // e.g [test](//test.com) - not "auto link"
+            // e.g [test.com](//test.com) - "auto link"
+            const auto =
+              node.hasAttribute(AUTO_ATTRIBUTE) ||
+              href === text ||
+              href?.replace(`${this.options.defaultProtocol}//`, '') === text;
```
whawker, line 301: Think we should also check `this.options.autoLink` is enabled for the new checks here?

marcoSven: If the autoLink option is not set to true does it matter if that value is true or false?

Should I change this to 

```ts 
const auto =
this.options.autoLink &&
(node.hasAttribute(AUTO_ATTRIBUTE) ||
href === text ||
href?.replace(`${this.options.defaultProtocol}//`, "") ===
text);
```

whawker: > If the autoLink option is not set to true does it matter if that value is true or false?

I just want to avoid the potential confusion of people seeing `auto: true` in there doc state when the think they have auto linking disabled.

Your suggested code makes sense to me

marcoSven: Thank you for the explanation.

Pushing the change.

whawker:
Thanks so much for the contribution here! Much appreciated!

PR #1737, opened 2022-06-22
feature(ext-link): optional link validation function

marcoSven:
### Description
Ability to provide a custom link validation function for the Remirror link extension.

Opening as a draft, this feature is based on a branch with an open pull request #1743

closes #1730

### Checklist

- [x] I have read the [**contributing**](https://github.com/remirror/remirror/blob/HEAD/docs/contributing.md) document.
- [x] My code follows the code style of this project and `pnpm fix` completed successfully.
- [x] I have updated the documentation where necessary.
- [x] New code is unit tested and all current tests pass when running `pnpm test`.

PR #1743, opened 2022-06-24
fix(ext-link): not auto detecting adjacent text

### LINKED ISSUE #1732: Links not updating based on adjacent text nodes - link extension
### Summary

Links should detect trailing characters that invalidate a previously valid link.

### Steps to reproduce

https://remirror.io/

1. Type "window.co" - creates a valid link
2. Continue typing "nfirm"

### Expected results

```html
<p>
  windo.confirm
</p>
```

### Actual results

```html
<p>
  <a href="//windo.co">windo.co</a>
  nfirm
</p>
```

### Screenshot(s)

https://user-images.githubusercontent.com/3860298/174818160-8eb39d9f-23b2-4b7f-9594-6afc8ec1300e.mov

marcoSven:
#### Before this change:

Auto link detection did not detect adjacent text.

**Example:**

Insert "window.confirm" results in "[window.co](//window.co)nfirm"

#### After this change:

Adjacent text is detected and possibly removes invalid link or creates a link

**Example**:

Insert "window.confirm" results in "window.confirm"

Added "adjacentPunctuations" option to configure which adjacent punctuation are excluded from links

Closes #1732

### Checklist

### Screenshots

https://user-images.githubusercontent.com/3860298/174903466-bcac9fea-e4cb-4233-989f-7c87600b1659.mov


marcoSven:
@whawker we have now 3 different conflicting pull requests https://github.com/remirror/remirror/pull/1734, https://github.com/remirror/remirror/pull/1774, and this pr https://github.com/remirror/remirror/pull/1743 :)

#1774 #1743 have a similar UX which I favor over creating a link after a trigger character.

I just did a refactor that simplifies and reduces the complexity of this change. Again one of the motivations of this pr is the ability to easily add the ability to provide a link validation method. Would it help if I update https://github.com/remirror/remirror/pull/1737?

Happy to take feedback and improve this pr.

whawker:
Thanks for the refactor, I much prefer the reduced complexity!

Happy to provide feedback here, I just wanted to offer an alternative with a reduced changeset

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -603,29 +665,84 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
             }
 
             const nodeText = tr.doc.textBetween(pos, pos + node.nodeSize, undefined, ' ');
-            this.findAutoLinks(nodeText)
+
+            const foundAutoLinks: FoundAutoLink[] = [];
+
+            for (const match of findMatches(nodeText, new RegExp(/(\S)+/, 'gm'))) {
```
whawker, line 671: Can we stored this regex in a const at the start of this file? It gives a slight performance benefit (unless things have changed) - which is valuable as this is executed quite frequently

```js
const WHITESPACE_REGEX = /\S+/gm;
```

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -603,29 +665,84 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
+                urlInMatchRange &&
+                (linkMarkInRange.to >= to ||
+                  !within(from, linkMarkInRange.from, linkMarkInRange.to)) &&
+                linkMarkInRange.mark.attrs.href === this.buildHref(urlInMatchRange)
+              ) {
+                continue;
+              }
+
+              // Skip if no valid link is in the match
+              if (!urlInMatchRange && !new RegExp('^[0-9|-]+$').test(matchedText)) {
```
whawker, line 697: Can you explain the intent of this regex here, I'm not sure what were asserting?

I suspect the vertical bar in the character group here is a mistake? See [here](https://regexper.com/#%5E%5B0-9%7C-%5D%2B%24)

P.s. we should also store this in a const

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -739,6 +835,91 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
+    text,
+  }: {
+    url?: string;
+    startIndex?: number;
+    text: string;
+  }) {
+    const newUrl = createNewURL(text);
+
+    // Check for trailing punctuation and adjacent characters
+    if (newUrl) {
```
whawker, line 865: Instead of wrapping the entire block, I would prefer exiting early

```suggestion
    if (!newUrl) {
      return;
    }
```

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -739,6 +835,91 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
+        input: text,
+        url,
+      });
+
+      const { pathname, search } = newUrl;
+
+      let decoded;
+
+      try {
+        decoded = decodeURI(pathname.slice(1) + search);
```
whawker, line 879: I'm not sure I see the value in decoding a URL to check for special characters?

In my mind if they are URL encoded, by definition they are part of the URL.

packages/remirror__extension-link/src/link-extension-utils.ts
```diff
@@ -53,3 +53,83 @@ export const TOP_50_TLDS = [
...
+  input: string;
+  url: string | undefined;
+}) => {
+  const length = ((url && input.split(url)[1]) || '').length;
+
+  if (length > 1) {
+    return Math.abs(length) * -1;
+  }
+
+  return;
```
whawker, line 121:
```suggestion
  if (!url) {
    return 0;
  }

  return input.split(url)[1].length * -1; // No need to worry about -0 as -0 === 0
```

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -739,6 +835,91 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
+  }: {
+    url?: string;
+    startIndex?: number;
+    text: string;
+  }) {
+    const newUrl = createNewURL(text);
+
+    // Check for trailing punctuation and adjacent characters
+    if (newUrl) {
+      const endsWithPunctuation = (input: string) =>
```
whawker, line 866: Think we could make life easier here, if this actually returned the end punctation, rather than just saying there is some

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -739,6 +835,91 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
+      try {
+        decoded = decodeURI(pathname.slice(1) + search);
+      } catch {
+        decoded = pathname.slice(1) + search;
+      }
+
+      if (decoded && endsWithPunctuation(decoded)) {
+        const index = -1;
+        const balancedIndex = !isBalanced(decoded) ? getBalancedIndex(decoded, index) : 0;
+        const trailingPunctuationIndex = SENTENCE_PUNCTUATIONS.includes(decoded.slice(-1))
```
whawker, line 887: No need to calculate this until line 900

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -739,6 +835,91 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
+      const urlMatch = getMatchString(match);
+
+      if (!this.isValidUrl(urlMatch)) {
+        continue;
+      }
+
+      return urlMatch;
+    }
+
+    return;
```
whawker, line 850: Loop might be unnecessary here - we have already split the text on whitespace characters, so there should only be one link within each sub string?

```suggestion
    const matches = findMatches(input, this.options.autoLinkRegex);
    
    if (!matches || !this.isValidUrl(match[0])) {
      return;
    }
    
    return matches[0];
```

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -739,6 +835,91 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
+      if (adjacentCharCount && !this.isValidUrl(text.slice(startIndex))) {
+        return (
+          adjacentCharCount +
+          // Adjust index when switching to longer TLD e.g '.co'  to '.com'
+          (this.isValidUrl(text.slice(startIndex, adjacentCharCount + 1)) ? 1 : 0)
+        );
+      }
+
+      if (endsWithPunctuation(text)) {
+        return -1;
```
whawker, line 917: Think we're assuming here that is ends with a single punctuation mark? Part of the reason why I think `endsWithPunctation` should return the string instead?

packages/remirror__extension-link/src/link-extension.ts
marcoSven: Good call

packages/remirror__extension-link/src/link-extension.ts
marcoSven: This was from a previous iteration, checking for phone numbers. Thanks for calling it out - consider it removed

packages/remirror__extension-link/src/link-extension.ts
marcoSven: Absolutely was meaning to that

marcoSven: Absolutely was meaning to do that

packages/remirror__extension-link/src/link-extension.ts
marcoSven: Yes I should have done that

packages/remirror__extension-link/src/link-extension.ts
marcoSven: Agree, let me take a look

packages/remirror__extension-link/src/link-extension.ts
marcoSven: That is a good point

packages/remirror__extension-link/src/link-extension.ts
marcoSven: good catch

packages/remirror__extension-link/src/link-extension-utils.ts
```diff
@@ -53,3 +53,83 @@ export const TOP_50_TLDS = [
...
+};
+
+export const addProtocol = (input: string) =>
+  ['http://', 'https://', 'ftp://'].some((protocol) => input.startsWith(protocol))
+    ? input
+    : `https://${input}`;
```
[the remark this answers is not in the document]
marcoSven: Done

packages/remirror__extension-link/src/link-extension.ts
marcoSven: Think this is resolved, otherwise let me know, please.

marcoSven:
Thank you for taking the time and the excellent feedback @whawker. I addressed most of your requests and marked them resolved. Please let me know if I missed something.

packages/remirror__extension-link/src/link-extension.ts
marcoSven: Let me just remove all of that we don't need to use `URL()` just to get the path.

whawker:
I'm seeing a few regressions on this branch, I'm trying to track down the cause, but I'd appreciate some assistance.

1. It seems to be creating links on text that I'm not currently editing. 

    https://user-images.githubusercontent.com/2003804/178735509-9fd5003e-e5da-4bb4-b60b-7049e3e0d2ea.mov

    FYI This should be avoided as it makes it impossible to _remove_ a link, as the auto link behaviour adds it straight back again.

2. Additionally, it seems like when adding the same link twice, it takes a while to be recognised - then when it is detected it removes the previous link instance (not the first line, when the link is added on the second line)

    https://user-images.githubusercontent.com/2003804/178735969-fb0b57f7-8ceb-4b2b-b3f8-201a21b0309b.mov

marcoSven:
@whawker I was able to reproduce the first issue in our project - thanks for catching it.

Also addressed the second issue.

I will add tests but want to confirm they are fixed + your feedback if any. Also, let me know if you can find anything else - Thank you

whawker (changes requested):
I've been trying to wrap my head around this new auto link logic for a few days now and I can't make sense of how this new logic works. 

I can see that we loop through each node and split on whitespace to try and find links within each piece, but then I find myself get thoroughly confused.

I can see that were working out a negative index to slice unnecessary characters from the end of the string, however when I try to run the code through in my head I keep finding myself looking at multiple strings with [negative index slices](https://github.com/remirror/remirror/blob/b1058d93d57fe3821a5f7f4b36f6a350a732e390/packages/remirror__extension-link/src/link-extension.ts#L841-L870), along with a [performing splits](https://github.com/remirror/remirror/blob/b1058d93d57fe3821a5f7f4b36f6a350a732e390/packages/remirror__extension-link/src/link-extension-utils.ts#L105) to trim strings.

In its current state I can't accept this PR, as I don't think I'll be able to maintain this code going forward, as I'm really struggling to understand it.

My gut feel is to resurrect #1774, and address the issues you raised on that PR instead.

marcoSven:
I hear you @whawker closing this pr in favor of #1793 


I added some pr comments that maybe help to better understand what is happening and my reasoning behind it. Happy to clarify or make further adjustments.

PR #1756, opened 2022-06-28
fix(link): delay trigger of onUpdateLink event to prevent stale state updates

### LINKED ISSUE #1754: link ext onUpdateLink DOMException - Failed to execute 'collapse' on 'Selection'
### Summary

The issue is with the "onUpdateLink" handler when used with Remirror context. The context has still the previous state. Not 100% sure if this is a bug or if that is intentional. 

### Steps to reproduce

See code sandbox https://codesandbox.io/s/remirror-update-link-u729f1?file=/src/App.tsx
Enter "test.com" and backspace ( delete ) the "m" - "test.co"

### Expected results

no error

### Actual results

"Failed to execute 'collapse' on 'Selection': The offset 8 is larger than the node's length (7)."

### Possible Solution

Push the "onUpdateLink" call to the end of the execution queue

### Screenshot(s)

https://user-images.githubusercontent.com/3860298/176131214-eff5c861-6c27-4cbd-b659-aab6cd7c8d77.mov

marcoSven:
### Description

closes #1754

### Checklist

### Screenshots

see issue #1754

### DIFF
--- a/packages/remirror__extension-link/src/link-extension.ts
+++ b/packages/remirror__extension-link/src/link-extension.ts
@@ -624,7 +624,11 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
 
           onUpdateCallbacks.forEach(({ range, attrs, text }) => {
             const { doc, selection } = tr;
-            this.options.onUpdateLink(text, { doc, selection, range, attrs });
+
+            // Push to the end of the execution queue to prevent possible DOMException
+            setTimeout(() => {
+              this.options.onUpdateLink(text, { doc, selection, range, attrs });
+            }, 0);
           });
         });
 
--- a/packages/remirror__extension-link/src/link-extension.ts
+++ b/packages/remirror__extension-link/src/link-extension.ts
@@ -622,13 +622,11 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
               });
           });
 
-          onUpdateCallbacks.forEach(({ range, attrs, text }) => {
-            const { doc, selection } = tr;
-
-            // Push to the end of the execution queue to prevent possible DOMException
-            setTimeout(() => {
-              this.options.onUpdateLink(text, { doc, selection, range, attrs });
-            }, 0);
+          window.requestAnimationFrame(() => {
+            onUpdateCallbacks.forEach(({ attrs, range, text }) => {
+              const { doc, selection } = tr;
+              this.options.onUpdateLink(text, { attrs, doc, range, selection });
+            });
           });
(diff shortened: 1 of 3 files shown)


packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -624,7 +624,11 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
 
           onUpdateCallbacks.forEach(({ range, attrs, text }) => {
             const { doc, selection } = tr;
-            this.options.onUpdateLink(text, { doc, selection, range, attrs });
+
+            // Push to the end of the execution queue to prevent possible DOMException
+            setTimeout(() => {
```
whawker, line 629: Did you try any alternatives here? Such as `requestAnimationFrame`?

marcoSven: I have not as I mentioned in the [issue](https://github.com/remirror/remirror/issues/1754#issue-1287020628) I am not even sure if this is a bug or intentional. 

Don't think `requestAnimationFrame` would work though, have you looked at the code sandbox?

whawker: > Don't think `requestAnimationFrame` would work though

Looks like it does, you're correct that this is being fired at the wrong point in time though. It means no commands can be trigger from in the event handler, as it is executed against stale state.

P.s. I would rather not delay in a loop, I  would recommend wrapping the whole block instead recommend this was changed to

```ts
window.requestAnimationFrame(() => {
  onUpdateCallbacks.forEach(({range, attrs, text}) => {
    const {doc, selection} = tr;
    this.options.onUpdateLink(text, {doc, selection, range, attrs});
  });
});
```

marcoSven: >  I would recommend you change your code to use the dedicated hooks, rather than using the manager directly.


Thank you for the recommendation.


> Looks like it does, you're correct that this is being fired at the wrong point in time though. It means no commands can be executed within the event handler, as it is executed against stale state.

Great let me push a commit

whawker:
Thanks so much again @marcoSven!

marcoSven:
Thank you @whawker, glad it turned out to be a valid bug that's fixed now.

Going to fix the new conflicts and failing tests in this pr #1743 because of this merge. 

Have you had a chance to look at it yet? I realize it's a larger change. 
We actually already merged the changes into our main branch. It does feel much more responsive to us. 

Looking forward to your feedback.

Issue #1730, opened 2022-06-21
Custom link validation function option - link extension


marcoSven:
### Description

Ability to provide a custom link validation function for the Remirror link extension.

### Context

For a project that already uses another method of link validation, it would be more maintainable and easier to match results.

Amethystafyy282:
I think this might help.
`new LinkExtension({ autoLinkRegex: 'your_regexp' }),`

marcoSven:
Thank you @Amethystafyy282 I am aware of this option but I am proposing the option to provide your own validation that would be used instead of the autoLinkRegex

https://github.com/marcoSven/remirror/commit/f7f824fd4e72e53e16f69b4617d8f1190009ef8a#diff-c6362a6560783faf0b832a4b10098233042a3092f87b84b29f28f46c94410ac9R158

PR #1774, opened 2022-07-07
fix(link): prevent trailing punctuation being include in auto link URL

whawker:
### Description

Similar to #1734 and #1743 - prevent including trailing punctation in an auto link URL.

Also prevent cases such as [window.co](https://window.co)nfirm

@marcoSven and @DiegoPayan I'd really appreciate your input here to see if this solves the issues you've been seeing.

I've opted for this approach to limit changes to the code.

### Checklist

### DIFF
--- a/packages/remirror__extension-link/__tests__/link-extension.spec.ts
+++ b/packages/remirror__extension-link/__tests__/link-extension.spec.ts
@@ -654,6 +654,10 @@ describe('autolinking', () => {
       },
       attrs: { auto: true, href: '//test.com' },
     });
+
+    onUpdateLink.mockClear();
+    editor.insertText(' ');
+    expect(onUpdateLink).not.toHaveBeenCalled();
   });
 
   it('should not autolink or call onUpdateLink by default', () => {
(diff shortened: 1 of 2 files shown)


packages/remirror__extension-link/__tests__/link-extension.spec.ts
```diff
@@ -909,9 +1132,6 @@ describe('more auto link cases', () => {
     // with "#"
     { input: "https://en.wikipedia.org/wiki/The_Power_of_the_Powerless#Havel's_greengrocer" },
 
-    // with '(' and ')'
-    { input: 'https://en.wikipedia.org/wiki/Specials_(Unicode_block)' },
```
marcoSven, line 913: Should not introduce a regression. This is a valid link.

marcoSven:
Thank you @whawker, on a quick glance this mostly solves it too.

What is missing for us is the following

- Ability to provide a URL validation method, we are using http://markdown-it.github.io/linkify-it/ to validate URLs. I know this is not part of this pr but I am looking for a path forward.
- Balanced pairs (braces) appear not to be working yet. A closing  ")" is not part of the URL.
- Configurable adjacent punctuations ( minor issue )

The procedural solution in this pr #1743 uses two fuzzy matches before checking the validity of a URL and trimming the path if needed.

The benefits are

1. The ability to use a custom URL validation method instead of the default  `autoLinkRegex` and TLDs. See #1737 and #1730.
2. Configurable adjacent punctuations that are excluded from a link. This is preferred over trying to adjust the `autoLinkRegex`.
3. Regardless of the `autoLinkRegex`, trailing punctuation and adjacent characters are handled based on link validation ( default `autoLinkRegex` + TLD or the proposed optional URL validation method). For example, the balanced pair issue I mentioned above that the change you are proposing is not addressing see this failing [test](https://github.com/remirror/remirror/pull/1774/files#diff-eb7309e3871018a97a8f5fef74d56d955163606e967e3c875dbe333b78887886L912).

marcoSven (changes requested):

PR #1775, opened 2022-07-07
fix(theme): wrap end-of-lines spaces

ocavue:
### Description

Don't set `white-space` as `pre-wrap`, because this will hang end-of-lines spaces, which make the content inside a `<p>` element have more width than the `<p>` element. 

At the top of `core-theme.ts`, we already have [`white-space: break-spaces;`](https://github.com/remirror/remirror/blob/remirror%402.0.0-beta.5/packages/remirror__theme/src/core-theme.ts#L11-L12
). So I just delete these override styles. I don't see the potential uses of these styles anyway. 

### Checklist

### Screenshots

https://user-images.githubusercontent.com/24715727/177761259-f8d93125-67d8-4d81-9e93-f096f2f545a7.mp4

### DIFF
--- a/packages/remirror__styles/all.css
+++ b/packages/remirror__styles/all.css
@@ -594,14 +594,6 @@ button:active .remirror-menu-pane-shortcut,
   position: relative;
 }
 
-.remirror-editor.ProseMirror[contenteditable='false'] {
-  white-space: normal;
-}
-
-.remirror-editor.ProseMirror[contenteditable='true'] {
-  white-space: pre-wrap;
-}
-
 .remirror-editor.ProseMirror hr {
   border-color: #2e2e2e;
 }
--- a/.changeset/seven-timers-prove.md
+++ b/.changeset/seven-timers-prove.md
@@ -1,0 +1,6 @@
+---
+'@remirror/styles': patch
+'@remirror/theme': patch
+---
+
+Set style `white-space` as `break-spaces` to wrap end-of-lines spaces.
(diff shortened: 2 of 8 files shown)


packages/remirror__styles/all.css
```diff
@@ -594,14 +594,6 @@ button:active .remirror-menu-pane-shortcut,
   position: relative;
 }
 
-.remirror-editor.ProseMirror[contenteditable='false'] {
-  white-space: normal;
-}
-
-.remirror-editor.ProseMirror[contenteditable='true'] {
-  white-space: pre-wrap;
-}
```
whawker, line 603: Possible issue here (though unlikely?) is that our browser support is `since 2017` ([these browsers](https://en.wikipedia.org/wiki/List_of_web_browsers#Historical)).

`white-space: break-spaces` isn't supported by [until more recent versions](https://caniuse.com/mdn-css_properties_white-space_break-spaces).

Perhaps we should just [copy ProseMirror here](https://github.com/ProseMirror/prosemirror-view/blob/aaa50d592074c8063fc2ef77907ab6d0373822fb/style/prosemirror.css#L7-L8)?

```css
.remirror-editor.ProseMirror {
  white-space: pre-wrap;
  white-space: break-spaces;
}
```

ocavue: We already have these two rules, right above the changed lines in this file.

whawker: Whoops! 🤦‍♂️

whawker (approved):

PR #1788, opened 2022-07-18
feat: add support for optional clickmark handler

sudhons:
This helps to pass an optional callback to handle clicks on highlight marks

### Checklist

### DIFF
--- a/packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
+++ b/packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
@@ -20,6 +20,7 @@ import {
   uniqueId,
   within,
 } from '@remirror/core';
+import type { CreateEventHandlers } from '@remirror/extension-events';
 import { Node } from '@remirror/pm/model';
 import { EditorStateConfig, TextSelection } from '@remirror/pm/state';
 import { DecorationSet } from '@remirror/pm/view';
@@ -101,6 +102,37 @@ export class EntityReferenceExtension extends MarkExtension<EntityReferenceOptio
         },
         ...(override.parseDOM ?? []),
       ],
+    };
+  }
+
+  /**
+   * Track click events passed through to the editor.
+   */
+  createEventHandlers(): CreateEventHandlers {
+    return {
+      /**
+       * listens to click events and call the "onClickMark" handler with any of:
+       * 1. no argument if the text clicked is not an highlight mark
+       * 2. the id of the clicked highlight mark
+       * 3. id of the shortest highlight mark in case of overlapping highlights
+       */
+      clickMark: (_event, clickState) => {
+        const { markRanges } = clickState;
+
+        if (markRanges.length === 0) {
+          return this.options.onClickMark();
+        }
+
+        let shortestMark = markRanges[0];
+
+        markRanges.forEach((mark) => {
+          if (mark.to - mark.from < shortestMark.to - shortestMark.from) {
+            shortestMark = mark;
+          }
+        });
+
+        return this.options.onClickMark(shortestMark.mark.attrs.id);
+      },
     };
   }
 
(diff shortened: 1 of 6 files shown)


packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -62,6 +62,7 @@ const createDecorationSet = (props: StateProps) => {
   defaultOptions: {
     getStyle: decorateEntityReferences,
     blockSeparator: undefined,
+    onClickMark: () => {},
```
ronnyroeller, line 65: Could you pls add a unit test for this? I'm aware that this is testing likely the functionality of an underlying classes. Yet, I find it important because the users of the `EntityReferenceExtension` might rely on this functionality - and they don't care what provides it.

See here an example: https://github.com/remirror/remirror/blob/bb33bd813fe05f8d9feea21eb75fd7bb20ac8b63/packages/remirror__extension-link/__tests__/link-extension.spec.ts#L1042

packages/remirror__extension-entity-reference/src/types.ts
```diff
@@ -24,6 +24,7 @@ export interface EntityReferenceOptions {
    */
   getStyle?: (entityReferences: EntityReferenceMetaData[][]) => Decoration[];
   blockSeparator?: AcceptUndefined<string>;
+  onClickMark?: (id: string) => void;
```
ronnyroeller, line 27: Could you pls add a short comment how this works. My first reaction was: “Why do we declare a handler but never use it?”
I guess that’s because there is some underlying class that actually does something with it - and our options before simple.
What might be even better to bring the point across: extend `EntityReferenceOptions` from the option of this underlying class (just picking `onClickMark`). This would communicate very clearer to the reader of the code why the handler is there.

ronnyroeller (changes requested):

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -103,6 +105,37 @@ export class EntityReferenceExtension extends MarkExtension<EntityReferenceOptio
...
+        const { markRanges } = clickState;
+
+        if (markRanges.length === 0) {
+          return this.options.onClickMark();
+        }
+
+        let shortestMark = markRanges[0];
+
+        markRanges.forEach((mark) => {
+          if (mark.to - mark.from < shortestMark.to - shortestMark.from) {
```
ronnyroeller, line 129: Can we reuse here the [getShortestEntityReference](https://github.com/remirror/remirror/blob/main/packages/remirror__extension-entity-reference/src/utils/shortest-entity-reference.ts)?

ronnyroeller (commented):
Pls see comment

packages/remirror__extension-entity-reference/src/entity-reference-extension.ts
```diff
@@ -103,6 +105,37 @@ export class EntityReferenceExtension extends MarkExtension<EntityReferenceOptio
...
+  createEventHandlers(): CreateEventHandlers {
+    return {
+      /**
+       * listens to click events and call the "onClickMark" handler with any of:
+       * 1. no argument if the text clicked is not an highlight mark
+       * 2. the id of the clicked highlight mark
+       * 3. id of the shortest highlight mark in case of overlapping highlights
+       */
+      clickMark: (_event, clickState) => {
+        const { markRanges } = clickState;
```
IdrissMahjoubi, line 120: From what I know this returns all mark ranges, do we need to check if the mark type is actually "entity-reference" ?

ronnyroeller (changes requested):
Code looks good.

Could you pls add a changeset (we use this to create new versions of Remirror)?
You can create a changeset via `pnpm changeset`. This should be a minor change to the entity-reference-extension (as it adds functionality).

ronnyroeller (approved):
LGTM

PR #1794, opened 2022-07-20
fix(entity-reference): fix handling overlapping marks on click event

IdrissMahjoubi:
### Description

This gives more flexibility how to react to a click, e.g. show all entity - instead of just the shortest which is a specific use case.
The `onClick` callback receives the mark attributes, it's range and it's text as opposed to just the attributes
Fix handling of overlapping marks on click position

Note: _This PR causes an API change. Yet, it makes the API more flexible and should be OK considering that we're still in v2-beta._

### Checklist

### DIFF
--- a/.changeset/gorgeous-bulldogs-look.md
+++ b/.changeset/gorgeous-bulldogs-look.md
@@ -1,0 +1,5 @@
+---
+'@remirror/extension-entity-reference': minor
+---
+
+Update pnpm-lock.yaml


ronnyroeller (changes requested):
Code looks good to me.

Changeset is also good: "Click event receives all entity reference marks, their ranges and their respective text on the clicked position".
I don't understand the PR description: "The extension uses click event instead of clickMark
The onClick callback receives the mark attributes, it's range and it's text as opposed to just the attributes. Fix handling of overlapping marks on click position"
=> Suggestion: Use the same as the changset, add why it was done (e.g. _"This gives more flexibility how to react to a click, e.g. show all entity - instead of just the shortest"_), and add a note on changing API (e.g. _"This PR causes an API change. Yet, it makes the API more flexible and should be OK considering that we're still in v2-beta."_)

PR #1799, opened 2022-07-25
Add helper for text color extention

Anenth:
### Description

Added a helper to get the color from the current selection 

### Checklist

### DIFF
--- a/packages/remirror__extension-text-color/src/text-color-extension.ts
+++ b/packages/remirror__extension-text-color/src/text-color-extension.ts
@@ -116,6 +116,18 @@ export class TextColorExtension extends MarkExtension<TextColorOptions> {
   }
 
   /**
+   * Get the color for the selection
+   */
+  @helper()
+  getTextColorForSelection(selection?: Selection): Helper<string> {
+    if (!selection) return this.options.defaultColor;
+
+    const state = this.store.getState();
+    const textMarkInfo = getMarkRange(selection.$from, this.type);
+    if (!textMarkInfo) return  this.options.defaultColor;
+    return textMarkInfo.mark.attrs.color;
+  }
+  /**
    * Get the color from the provided string. The string can be a computed
    * property as well.
    */


packages/remirror__extension-text-color/src/text-color-extension.ts
```diff
@@ -115,6 +115,18 @@ export class TextColorExtension extends MarkExtension<TextColorOptions> {
     return this.store.commands.removeMark.original({ type: this.type, ...options, expand: true });
   }
 
+  /**
+   * Get the color for the selection
+   */
+  @helper()
+  getTextColorForSelection(selection?: Selection): Helper<string> {
```
whawker, line 122: We should default to the current selection here, I believe

```ts
import { Selection } from '@remirror/core';
```

```suggestion
  getTextColorForSelection(selection: Selection = this.store.getState().selection): Helper<string> {
```

packages/remirror__extension-text-color/src/text-color-extension.ts
```diff
@@ -115,6 +115,18 @@ export class TextColorExtension extends MarkExtension<TextColorOptions> {
...
+    if (!selection) return this.options.defaultColor;
```
whawker, line 123: I suspect we need to check we have a text selection here?

```ts
import { isTextSelection } from '@remirror/core';
```

```suggestion
    if (!isTextSelection(selection) || selection.empty) {
      return this.options.defaultColor;
    }
```

packages/remirror__extension-text-color/src/text-color-extension.ts
```diff
@@ -115,6 +115,18 @@ export class TextColorExtension extends MarkExtension<TextColorOptions> {
...
+
+    const state = this.store.getState();
```
whawker, line 125: This doesn't seem to be used?

packages/remirror__extension-text-color/src/text-color-extension.ts
```diff
@@ -115,6 +115,18 @@ export class TextColorExtension extends MarkExtension<TextColorOptions> {
...
+    const textMarkInfo = getMarkRange(selection.$from, this.type);
+    if (!textMarkInfo) return  this.options.defaultColor;
```
whawker, line 127: Formatting is a bit off here, suspect this would be fixed by running `pnpm fix`

```suggestion
    
    if (!textMarkInfo) {
      return this.options.defaultColor;
    }
    
```

whawker (commented):
Thanks so much for the contribution @Anenth!

Sorry my review is so late, my normal job has been taking up all my time recently

PR #1800, opened 2022-07-25
fix(ext-link): not auto detecting adjacent text

### LINKED ISSUE #1732: Links not updating based on adjacent text nodes - link extension

marcoSven:
#### Before this change:

Auto link detection did not detect adjacent text.

**Example:**

Insert "window.confirm" results in "[window.co](//window.co)nfirm"

#### After this change:

Adjacent text is detected and possibly removes invalid link or creates a link

**Example**:

Insert "window.confirm" results in "window.confirm"

**Valid link examples:** 

- (((([www.remirror.io/((test))](www.remirror.io/((test)))))))
- [www.remirror.io/((test))](www.remirror.io/((test)))
- [www.remirror.io/((test)test](www.remirror.io/\(\(test\)test)
- [www.remirror.io/((test)](www.remirror.io/\(\(test\))
- [www.remirror.io/((test](www.remirror.io/\(\(test)
- "[www.remirror.io/test"""test](www.remirror.io/test"""test)"
- [www.remirror.io/test](www.remirror.io/test).
- [www.remirror.io/test](www.remirror.io/test)?
- [www.remirror.io/test?test=test](www.remirror.io/test?test=test)?

Added "adjacentPunctuations" option to configure which adjacent punctuation are excluded from links

Closes #1732

### Checklist

### Screenshots

https://user-images.githubusercontent.com/3860298/174903466-bcac9fea-e4cb-4233-989f-7c87600b1659.mov


whawker:
Can I ask - what is the use case for all the "balanced bracket pairs" logic?

It seems to add a whole load of complexity here, for a character that is very rarely used in a URL. I'm not sure I see the value

packages/remirror__extension-link/src/link-extension-utils.ts
```diff
@@ -53,3 +53,59 @@ export const TOP_50_TLDS = [
...
+    if (stack.length === 0 || stack.pop() !== index) {
+      return false;
+    }
+  }
+
+  return stack.length === 0;
+};
+
+export const getBalancedIndex = (input: string, index: number): number =>
+  !isBalanced(input.slice(0, index)) ? getBalancedIndex(input, --index) : index;
```
whawker, line 89: I suspect this logic could be simplified by reversing it - instead of recursively calling `isBalanced`, work out where the imbalance occurs on the original string

FYI, this returns a positive index - from the beginning of the string - rather than a negative one from the end

```ts
export const getBalancedIndex = (str: string): number => {
  const stack: Array<{ index: number, expected: string }> = [];

  for (let i = 0; i < str.length; i++) {
    const char = str.charAt(i);
    const bracketIndex = PAIR_PUNCTUATIONS.indexOf(char);

    // char is not a bracket
    if (bracketIndex === -1) {
      continue;
    }


    // char is an open bracket character
    if (bracketIndex % 2 === 0) {
      stack.push({ index: i, expected: PAIR_PUNCTUATIONS.charAt(bracketIndex + 1) });
      continue;
    }

    // char is a closing bracket character
    const unmatched = stack.pop();

    if (unmatched === undefined) {
      // closing bracket without any opening bracket
      return i;
    }

    const { index, expected } = unmatched;

    if (char !== expected) {
      // closing bracket was not what we expected
      return index;
    }
  }

  const unmatched = stack.pop();

  if (unmatched === undefined) {
    // str is balanced
    return -1;
  }

  return unmatched.index;
}

export const isBalanced = (str: string): boolean => {
  return getBalancedIndex(str) === -1;
};
```

packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -739,6 +803,55 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
+    // Balanced part of path
+    const balanced = path.slice(0, balancedPathIndex);
+    // Imbalanced part of path
+    const imbalanced = path.replace(balanced, '');
+
+    const balanceIndex = ![
+      ...imbalanced.slice((!isBalanced(imbalanced) ? getBalancedIndex(imbalanced, index) : 0) + 1),
+    ].some((char) => ![')', ']', '}'].includes(char) && !SENTENCE_PUNCTUATIONS.includes(char))
+      ? balancedPathIndex
+      : undefined;
```
whawker, line 846: I'm struggling to understand whats happening here

packages/remirror__extension-link/src/link-extension-utils.ts
marcoSven: Don't see the benefit of getting a positive index. Also, I am not sure if this will return the expected index without additional changes. I will test it though.

packages/remirror__extension-link/src/link-extension.ts
marcoSven: Yes sorry, let me add some comments. Also, think I can optimize this a little more.

marcoSven:
> Can I ask - what is the use case for all the "balanced bracket pairs" logic?
> 
> It seems to add a whole load of complexity here, for a character that is very rarely used in a URL. I'm not sure I see the value

I hear you and that is a very good point, I am simply addressing the edge cases here to provide a great UX. Pair punctuations can be part of the link or surrounding the link.

We essentially check once if a path is balanced. If the path is not balanced we are making an assumption at what point the URL ends so we can determine where to slice of trailing punctuations. 

For example, let's assume the user wants to wrap this balanced link in braces

[www.remirror.io/(test)](www.remirror.io/\(test\))

we are making the assumption that the link ended here ".../(test)"

([www.remirror.io/(test)](www.remirror.io/\(test\)))

@whawker if you feel very strongly about it and this is blocking the pr, could we make this optional?

But keep in mind if we remove this logic or disable it we would not know if a trailing closing brace is part of the URL.

[www.remirror.io/(test)](www.remirror.io/\(test\))

[www.remirror.io/(test](www.remirror.io/\(test))

Both are valid URLs but I would say the first is more likely the correct URL.

And when I paste this plain text `(www.remirror.io/(test))`

I would expect ([www.remirror.io/(test)](www.remirror.io/\(test\))) and not ([www.remirror.io/(test))](www.remirror.io/\(test\)\)).

packages/remirror__extension-link/src/link-extension-utils.ts
marcoSven: If this is a blocker for you I am happy to investigate further but as of right now it is not always returning the correct index.

marcoSven: @whawker I ended up implementing your suggestion after fixing the issue I noticed with it.

I was able to remove `isBalanced` and reduce the complexity in `getLinkEndIndex`.

PR #1812, opened 2022-08-02
feat(link): improve autolink behaviour with adjacent text

whawker:
### Description

cc @marcoSven please could you provide thoughts on this implementation.

Prevent trailing punctuation being included in the the autolink URL.

Remove a link, if a URL becomes invalid i.e. `window.co` -> `window.confirm`

### Checklist

### DIFF
--- a/packages/remirror__extension-link/src/link-extension.ts
+++ b/packages/remirror__extension-link/src/link-extension.ts
@@ -15,6 +15,7 @@ import {
   GetMarkRange,
   getMarkRange,
   getMatchString,
+  getSelectedGroup,
   getSelectedWord,
   Handler,
   includes,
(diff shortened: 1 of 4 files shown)


packages/remirror__extension-link/src/link-extension.ts
```diff
@@ -576,18 +577,38 @@ export class LinkExtension extends MarkExtension<LinkOptions> {
...
+            const newMarks = this.getLinkMarksInRange(doc, prevMark.from, prevMark.to, true);
 
             newMarks.forEach((newMark) => {
-              const wasLink = this._autoLinkRegexNonGlobal?.test(prevMark.text);
-              const isLink = this._autoLinkRegexNonGlobal?.test(newMark.text);
+              const wasLink = this.isValidUrl(prevMark.text);
+
+              // Expanded the link text to include additional non whitespace characters
+              // This determines if a link became invalid window.co -> window.confirm
+              const expanded = getSelectedGroup(state, /\s/, newMark) ?? newMark;
```
marcoSven, line 606: That's a nice solution

marcoSven (commented):
I can see the benefits of this implementation. The issue I have with this one-pass regex approach is that it will be harder to configure. 

Looking at nested braces, for example, maybe it is not a big deal for most, but it is frustrating if you need it. 

I do have some ideas on making this implementation easier to configure in the future, maybe this will be even easier and will not add complexity to the extension. 

I will check it out more later, I do think I see a regression. Do you mind if I make a pr into this branch with tests if I find regressions/bugs?

marcoSven:
Here is a similar regression that you pointed out on one of my earlier attempts. 

https://user-images.githubusercontent.com/3860298/183300702-ffad3a20-ba27-47b1-8b3c-62d2dd89df63.mov


https://user-images.githubusercontent.com/3860298/183300704-897c2328-b9b6-4d25-af46-bb925981e7d2.mov

whawker:
Thanks @marcoSven, 

By the way, the first video doesn't seem to work?

marcoSven:
The video was just very short, updated the video. The issue is essentially the same as in the second video. I addressed this regression in #1817

PR #1880, opened 2022-09-24
wip: polymorphic components test

whawker:
### Description

Experimenting reducing code duplication with polymorphic components

### Checklist


ocavue:
The code generation part looks good to me. 

For faster build speed, we can remove the prebuild (Prettier) script and just ignore these generated files from Prettier (and/or ESLint). 

(Idea: we can create our own file name pattern for generated files, like `xxx.gen.ts`. Then we can just add `*.gen.ts` into `.eslintignore` / `.prettierignore`)

PR #1892, opened 2022-09-30
fix(core-utils): don't throw error if window doesn't exist

ocavue:
### Description

Make `getStyle` and `getFontSize` robuster in server env. 

We shouldn't directly use the global variable `window` before checking the existing of it, otherwise, we could get the `ReferenceError: window is not defined` error in the server environment. 

### Checklist

### DIFF
--- a/packages/remirror__core-utils/src/core-utils.ts
+++ b/packages/remirror__core-utils/src/core-utils.ts
@@ -1237,16 +1237,12 @@ export function getDocument(): Document {
 }
 
 /**
- * Try to retrieve the window from the given HTML element or the global scope.
- *
  * @internal
  */
-export function maybeGetWindow(
-  element?: Element | null,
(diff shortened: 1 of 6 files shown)


packages/remirror__core-utils/src/core-utils.ts
```diff
@@ -1227,6 +1227,39 @@ export function getDocument(): Document {
...
+  );
+}
+
+/**
+ * @internal
+ */
+export function getWindow(
+  element?: Element | null,
+  document?: Document | null,
+): Window & typeof globalThis {
```
whawker, line 1253: Seems this is never called with both arguments.

Should this be two separate functions that wrap maybeGetWindow?

ocavue: Good catch. Done.

whawker (approved):

PR #1893, opened 2022-09-30
Export default mappings for React Renderer and add li

samtgarson:
### Description

Currently, to add new types locally to the React Renderer, we lose all the default mappings. This change allows us to import the default mappings to merge with any custom ones.

(This felt less intrusive than merging the types with the default ones by default)

It also adds a type registration for `li`.

### Checklist

### DIFF
--- a/packages/remirror__react-renderer/src/renderer.tsx
+++ b/packages/remirror__react-renderer/src/renderer.tsx
@@ -21,7 +21,7 @@ export const Doc: FC<SubRenderTreeProps> = ({ node, ...props }) => {
   return <div {...(node.attrs ?? object())}>{children}</div>;
 };
 
-const defaultTypeMap: MarkMap = {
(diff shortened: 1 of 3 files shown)


packages/remirror__react-renderer/src/renderer.tsx
```diff
@@ -35,9 +35,10 @@ const defaultTypeMap: MarkMap = {
   codeBlock: CodeBlock,
   orderedList: 'ol',
   text: TextHandler,
+  li: 'li'
```
ocavue, line 38: I guess The key here should be `listItem`?

samtgarson: Of course you’re right 🤦

samtgarson:
We’ve switched from using the React Renderer for now, so I’m not inclined to install pnpm and get this over the line, I’ll close for now.

Issue #1894, opened 2022-10-05
Retrieve the document from the global scope


FryingggPannn:
I’m trying to use Remirror with Next.js. Whenever I try to add an editor, using `<WysiwygEditor />` or `<Remirror />`, there’s an error saying
```
Error: Unable to retrieve the document from the global scope. Maybe you are running Remirror in a non-browser environment? If you are using Node.js, you can install JSDOM or similar to create a fake document and pass it to Remirror.
```

What does this mean? How can I resolve this issue?

I’m using the exact same code as shown in [the example](https://remirror.io/docs/getting-started/render-editor) (but without the CSS `import` at the top).

PR #1895, opened 2022-10-05
docs: clarify the error message from `getDocument`

ocavue:
### Description

Make the error message more detailed when Remirror is used on the server environment 

Close https://github.com/remirror/remirror/issues/1894 

### Checklist


ronnyroeller:
We could also consider to add the explanation to the documentation on remirror.io and then add a link to it in the error message (a la "This is the problem. For more details see: https://remirror.io/..."

PR #1896, opened 2022-10-06
fix(events): catch when `posFromDOM` returns -1

whawker:
### Description

`viewDes.posFromDom` [can return -1](https://github.com/ProseMirror/prosemirror-view/blob/master/src/viewdesc.ts#L248) which causes a RangeError to be thrown when we try to resolve the position.

### Checklist

### DIFF
--- a/.changeset/sweet-radios-relax.md
+++ b/.changeset/sweet-radios-relax.md
@@ -1,0 +1,5 @@
+---
+'@remirror/extension-events': patch
+---
+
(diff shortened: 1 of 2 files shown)


whawker:
@ocavue might be able to just use

```js
if (pos < 0) {
  return null;
}
```

But I'm away, and don't have my laptop with me right now

ocavue:
Sounds good. `posFromDOM` shouldn't return a number bigger than the document size anyway.

PR #1918, opened 2022-10-23
feat(find): add FindExtension and deprecate SearchExtension

ocavue:
### Description

This is a rewrite for the current search extension. I try to minimize the API and make it easier to understand. Also fixed some underlaying bug which causes the search result incorrect in a complex structure document. 

### Release plan 

I'd want to release a major version of `@remirror/extension-search` because its API is very different.

`remirror` and `@remirror/preset-wysiwyg` dependent on `@remirror/extension-search`. For these two packages, technically they have breaking change, but the change is related smaller (i.e. I don't think a lot of people are using `@remirror/extension-search`. So I plan to release a patch version for these two packages. 

Another approach would be creating a new NPM package (e.g. `@remirror/extension-find`) and deprecate `@remirror/extension-search`. This will not bring any breaking change, but I increase the maintaining workload. 
 
### React component `<FindReplaceComponent/>`

https://user-images.githubusercontent.com/24715727/197574501-83c9083f-50a8-4526-8c2f-1a1f6112c98d.mp4

### TODO: 

- [x] changeset 
- [x] docs
- [x] ~~submit a feature request to ProseMirror for `scrolltoSelection`, so that we can implement `scrollToActiveResult`~~ (we don't need new API from ProseMirror)
- [x] use MUI to implement a search & replace component in `@remirror/react-components` 
- [x] edit release script and make sure that releasing a major version of `@remirror/extension-search` won't release a major version of `remirror`. (maybe we can use [this](https://github.com/changesets/changesets/pull/542/files)) https://github.com/remirror/remirror/pull/1923

### Checklist


ocavue:
This PR is on draft, but I'd like to hear what do you think? @ronnyroeller @whawker

whawker:
You mentioned before this extension was broken - how broken? Did it not work at all?

If so, we can assume no one using it and avoid a major version bump?

ocavue:
> how broken? Did it not work at all?

All search result in the second and further list items won't get the correct position. 

CleanShot 2022-10-23 at 21 35 30@2x

ocavue:
Also, the existing extension won't work unless user write some custom CSS. (i.e. `@remirror/styles` doesn't provide the CSS search-extension needs to highlight the result)

ocavue:
> we can assume no one using it and avoid a major version bump

I prefer a major version bump for  `@remirror/extension-search` because (1) I cannot assume no one using it, and (2) It doesn't seems to be a big issue that one of our extensions has a higher major version than others. I think one of good part of releasing so many NPM packages is that we can only bring breaking to a small part of these packages and don't touch other packages.

packages/remirror__extension-search/src/search-extension.ts
```diff
@@ -306,86 +116,103 @@ export class SearchExtension extends PlainExtension<SearchOptions> {
...
+  createPlugin(): CreateExtensionPlugin {
+    return {
+      state: {
+        init() {
+          return DecorationSet.empty;
+        },
+        apply: (tr, old) => {
+          if (this._updating || (tr.docChanged && this.options.alwaysSearch)) {
+            const doc = tr.doc;
+            this._ranges = this.gatherSearchResults(doc);
```
whawker, line 153: Is searching the whole doc on each transaction expensive?

Could consider gathering the changed nodes and searching those?

ocavue: > Is searching the whole doc on each transaction expensive?

If the document is big, yes. 

> Could consider gathering the changed nodes and searching those?

That would be great. However I'd like to firstly finish the features (including the React search component) and then improve the performance. 

--- 

After a further research, I find that the browser's native search function is more similar to setting `alwaysSearch` as `false`. 

So I decide to set the default value of `alwaysSearch` as `false`. This will improve the default performance and give a not bad result.  

**Browser native search:**

https://user-images.githubusercontent.com/24715727/197514507-5bcb1289-1575-4f1b-89a6-a8281a828820.mp4


**`SearchExtension({alwaysSearch: false})`**:

https://user-images.githubusercontent.com/24715727/197514618-f58523b7-037e-4e60-9bf1-61f40a18ffd7.mp4


**`SearchExtension({alwaysSearch: true})`**:

https://user-images.githubusercontent.com/24715727/197514632-8ce31ffe-3498-4c36-9dab-40ba72281284.mp4

ocavue:
Note from last meeting:

- I will release a new package `@remirror/extension-find`, to avoid breaking change in the `remirror` package (`@remirror/extension-search` is exported by `remirror). 
- For the react component:
  - the find should be triggered after the typing (without the need of pressing Enter)
  - execute the search before replace

packages/remirror__react-components/src/find-replace/find-replace-component.tsx
```diff
@@ -0,0 +1,226 @@
...
+function useSearch() {
+  const helpers = useHelpers();
+  const commands = useCommands();
+  const [searchTerm, setSearchTerm] = React.useState('');
+  const [currIndex, setCurrIndex] = React.useState(-1);
+  const [total, setTotal] = React.useState(0);
+  const [caseSensitive, setCaseSensitive] = React.useState(false);
+  const [replacement, setReplacement] = React.useState('');
+
+  const search = (index: number) => {
```
IdrissMahjoubi, line 17: Maybe add `useCallback` to these functions to avoid recreating them on each rerender?

IdrissMahjoubi (approved):
LGTM (see comment)

PR #1948, opened 2022-11-25
fix(tables): apply table columns sizes when `editable=false`

### LINKED ISSUE #1937: Resized table columns is not reflected when editable is set to false
### Summary

We're using the Remirror editor to edit and display the content. After resizing the table columns and setting the editor mode to false, the table columns are back to equal width instead of the resized width.

### Steps to reproduce

Create an editor, resize the column, save the changes and display the content when the editor has the `editable` prop set to false.

### Expected results

Columns width should show the width set by the user

### Actual results

Columns width reverts back to equal width

### Possible Solution

N.A

### Screenshot(s)

Table issue

whawker:
### Description

Fixes #1937

Also adds a helper to the doc extension for determining if the doc content has been changed from the default

### Checklist

### DIFF
--- a/packages/remirror__extension-tables/src/table-extensions.ts
+++ b/packages/remirror__extension-tables/src/table-extensions.ts
@@ -4,6 +4,7 @@ import {
   CommandFunction,
   CommandFunctionProps,
   convertCommand,
+  CreateExtensionPlugin,
   EditorState,
   EditorView,
   extension,
@@ -12,6 +13,7 @@ import {
   findParentNodeOfType,
   Helper,
   helper,
+  isElementDomNode,
   NodeExtension,
   NodeSpecOverride,
   nonChainable,
@@ -38,9 +40,12 @@ import {
   setCellAttr,
   splitCell,
   tableEditing,
+  // @ts-expect-error TableView is exported
+  TableView,
   toggleHeaderCell,
   toggleHeaderColumn,
   toggleHeaderRow,
+  updateColumnsOnResize,
 } from '@remirror/pm/tables';
 
 import {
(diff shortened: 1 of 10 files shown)


packages/remirror__extension-tables/src/table-extensions.ts
```diff
@@ -38,9 +40,14 @@ import {
   setCellAttr,
   splitCell,
   tableEditing,
+  TableMap,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore it is exported
```
ocavue, line 45: let's use `// @ts-expect-error`

packages/remirror__extension-tables/src/table-extensions.ts
```diff
@@ -144,6 +149,43 @@ export class TableExtension extends NodeExtension<TableOptions> {
     return plugins;
   }
 
+  createPlugin(): CreateExtensionPlugin {
```
ocavue, line 152: Maybe we can directly use `createNodeViews` method here. The code will be simpler.

ocavue (approved):

packages/remirror__extension-tables/src/table-extensions.ts
whawker: We're using `updateExtensionPlugins` here (in the `onView` method),

This only checks the extension's `createPlugin` and `createExternalPlugins` methods.

[See here](https://github.com/remirror/remirror/blob/main/packages/remirror__core/src/builtins/plugins-extension.ts#L193-L216)

Issue #1993, opened 2023-02-11
YJS Extension editing interrupted by PlaceholderExtension decorations when collaborating


jeffrafter:
### Summary

When at least two users are collaborating (for example using the `WebrtcProvider`) and a user selects all and deletes the document becomes `empty`. This triggers the `PlaceholderExtension` which creates a decoration node for the placeholder message. Typing (collaborative) becomes disabled in the editing user window until they hit `enter`. This is because the node position / selection is confused and YJS can't sync.  

### Steps to reproduce

```ts
"use client";

import React from "react";
import { YjsExtension } from "@remirror/extension-yjs";
import { WebrtcProvider } from "y-webrtc";
import * as Y from "yjs";
import {
  Remirror,
  ThemeProvider,
  useRemirror,
} from "@remirror/react";

const room = "test-placeholder-yjs";
const ydoc = new Y.Doc();
const provider = new WebrtcProvider(room, ydoc);

const extensions = () => [
  new YjsExtension({ getProvider: () => provider, disableUndo: true })  
];

const RemirrorPage = (): JSX.Element => {
  const { manager } = useRemirror({
    extensions,
    core: { excludeExtensions: ["history"] },
  });

  return (
    <ThemeProvider>
      <Remirror manager={manager} autoFocus autoRender="end">
      </Remirror>
    </ThemeProvider>
  );
};

export default RemirrorPage;
```

### Expected results

After deleting the content of the document you should be able to immediately begin typing.

### Actual results

User can't type until they press enter or another user inserts content collaboratively.

### Possible Solution

1. Make it possible to remove the PlaceholderExtension from the ReactExtension core (`excludePlaceholder` or `core: { excludeExtensions: ["history", "placeholder"] }`). 
2. Create a new option on the `PlaceholderExtension` called `disabled`. If the extension is `disabled` do not apply the decorations. 
3. Implement the decoration as a Mark so it is shared (this seems very problematic and is probably not a good idea).

PR #1994, opened 2023-02-11
Allow disabling the placeholder

jeffrafter:
### Summary

When at least two users are collaborating (for example using the `WebrtcProvider`) and a user selects all and deletes the document becomes `empty`. This triggers the `PlaceholderExtension` which creates a decoration node for the placeholder message. Typing (collaborative) becomes disabled in the editing user window until they hit `enter`. This is because the node position / selection is confused and YJS can't sync.  

This allows you to disable the decorations in the placeholder. In general, I feel like this approach solves the issue (which is why I've opened this PR) but doesn't feel ergonomic. It would be better to remove the extension altogether in this case.

Fixes https://github.com/remirror/remirror/issues/1993

### DIFF
--- a/packages/remirror__extension-placeholder/src/placeholder-extension.ts
+++ b/packages/remirror__extension-placeholder/src/placeholder-extension.ts
@@ -23,6 +23,11 @@ export interface PlaceholderOptions {
    * then you will also need to apply your own styles.
    */
   emptyNodeClass?: string;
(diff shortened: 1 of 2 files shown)


andrew-maltsev:
hey any updates? this is really crucial bugfix for collaborating on mobile and desktop

PR #2011, opened 2023-03-16
chore(eslint): fix react version  setting

marcoSven:
### Description

React version. "detect" automatically picks the version you have installed.
You can also use `16.0`, `16.3`, etc, if you want to override the detected value.
It will default to "latest" and warn if missing, and to "detect" in the future

https://github.com/jsx-eslint/eslint-plugin-react/blob/f56851b433e7171991eb7e726f00596f14563c25/README.md?plain=1#L43

### Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -38,7 +38,7 @@ let config = {
   },
   settings: {
     react: {
-      version: 'latest',
+      version: 'detect',
     },
   },
   env: {


ocavue:
Thanks. Sorry I'm not completely following you, what's problem of current "latest" config?

marcoSven:
Not really a problem as far as I know, specifically when running `pnpm run lint:es`.

The reason for this pr is that "latest" is invalid, please see the [documentation](https://github.com/jsx-eslint/eslint-plugin-react/blob/f56851b433e7171991eb7e726f00596f14563c25/README.md?plain=1#L43).

I noticed it while looking into another unrelated issue when using "eslint_d" instead of "eslint" ( editor ). 


Screenshot 2023-03-17 at 06 38 58

PR #2017, opened 2023-04-01
fix(react-components): increase active index after replace

### LINKED ISSUE #2016: Replace isn't working correctly when trying to replace with similar word like `work` to `working`
### Summary

Replacing doesn't work if replacement has a subset of query.

### Steps to reproduce

* Go to find extension example -> https://remirror.io/docs/extensions/find-extension/#examples
* Search for 'dolor' and you will find four occurences
* Fill up the value `dolors` in the Replace input field
* Click `Replace` and it will always replace the first occurence
* Click `Next` to move the focus to the next word and Click `Replace` again and it will still replace the first occurence 

### Expected results

* It should ideally replace the occurence where the focus is at

### Actual results

* Replacing the first occurence all the time

### Possible Solution
* Replace the focused occurence

devcer:
### Description

Closes https://github.com/remirror/remirror/issues/2016

### Checklist

### DIFF
--- a/.changeset/blue-crabs-knock.md
+++ b/.changeset/blue-crabs-knock.md
@@ -1,0 +1,5 @@
+---
+'@remirror/react-components': patch
+---
+
+Increase active index after replace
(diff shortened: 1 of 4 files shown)


devcer:
@whawker not sure if this is the right way to do it but the bug can be fixed with this. Do let me know if any changes are needed.

ocavue (approved):

PR #2048, opened 2023-04-28
chore(lint): eslint parsing error tsconfig

marcoSven:
### Description

- set eslint parser project option to help eslint locate ts configurations 
- reduce lint warning noise, down to 9 warnings from prev 35 warnings.
- fix multiple lint issues that previously did not get caught - (we had a lot more than the 35 issues)
- fix "block statement surrounding arrow body" warnings, could revert that commit and disable the rule if we prefer that.
- md linting is intentionally disabled by default. Added `MD_ESLINT_CHECK` env. 

Edit:
Created a follow-up draft pr that enables md linting by default https://github.com/remirror/remirror/pull/2049. My personal pref is to disable it, I don't think it is worth it to lint markdown code blocks.

### Checklist

### Screenshots

lint-result


.eslintrc.js
```diff
@@ -35,6 +134,8 @@ let config = {
     ecmaFeatures: {
       jsx: true,
     },
+    project: 'tsconfig.eslint.json',
+    tsconfigRootDir: __dirname,
```
marcoSven, line 138: Relevant change is here

.eslintrc.js
```diff
@@ -495,25 +503,31 @@ if (process.env.FULL_ESLINT_CHECK) {
...
     // Apply the markdown plugin
     plugins: [...config.plugins, 'markdown'],
 
+    parserOptions: {
+      ecmaVersion: 2021,
+      sourceType: 'module',
+      ecmaFeatures: {
+        jsx: true,
+      },
+    },
```
marcoSven, line 518: Second relevant change. We could revert some of the other changes in this file, let me know.

ocavue: Why do we want a different `parserOptions` (i.e. not `parserOptions.tsconfigRootDir` and `parserOptions.project`) when `MD_ESLINT_CHECK` is true?

marcoSven: It's not possible to use type-aware linting rules with any ESLint processors.
Only @typescript-eslint rules that don't require a tsconfig.json are supported.

ocavue (approved):
Thank you!

PR #2088, opened 2023-06-08
fix(mention-atom): fix partial replacements including whitespace

whawker:
### Description

When using the Mention Atom extension, if you add a mention atom in the _middle_ of a paragraph, **and** your supported characters **allow for whitespace** (i.e. your labels are full names), you will find that selecting an option from the dropdown will greedily replace the remaining text in that line.

#### Before

https://github.com/remirror/remirror/assets/2003804/369578d8-07ad-4279-b140-09d4d19e6076

#### Changes

A `replacementType` [already exists in mention atom extension](https://github.com/remirror/remirror/blob/main/packages/remirror__extension-mention-atom/src/mention-atom-extension.ts#L311-L319), but it's behaviour has never been implemented. 

So I have implemented this behaviour in a similar fashion to the _Mention_ extension. This means that providing the `replacementType` option to the `useMentionAtom` hook, will now have an effect, replacing the text only up to the text cursor.

```tsx
const { state, getMenuProps, getItemProps, indexIsHovered, indexIsSelected } = useMentionAtom({
  items: users,
  replacementType: 'partial', // <-- Here
});
```

#### After

https://github.com/remirror/remirror/assets/2003804/cc30cd88-b630-44aa-9663-79cb5162055e

### Checklist


ocavue (approved):
LGTM. A small typescript issue needs to be fixed.

.changeset/moody-rules-suffer.md
```diff
@@ -0,0 +1,6 @@
+---
+'@remirror/extension-mention-atom': patch
+'@remirror/react-hooks': patch
```
ocavue, line 3: Let's add `@remirror/react` here too. I feel the number of published packages are usually too large, so I changed our publish strategy some while ago. Now only package that explicitly added here will be published

PR #2096, opened 2023-06-26
Fix/improve upload extension

shlroland:
### Checklist


packages/remirror__core/src/builtins/upload-extension/file-placeholder-actions.ts
```diff
@@ -15,4 +16,19 @@ interface RemovePlaceholderAction {
...
+export type PlaceholderPluginAction =
+  | AddPlaceholderAction
+  | RemovePlaceholderAction
+  | UpdatPlaceholderAction;
+
+export interface PlaceholderPluginMeta {
+  added: Array<Omit<AddPlaceholderAction, 'type'>>;
+  removed: Array<Omit<RemovePlaceholderAction, 'type'>>;
+  updated: Array<Omit<UpdatPlaceholderAction, 'type'>>;
+}
```
ocavue, line 34: If `PlaceholderPluginMeta` is just an array of `PlaceholderPluginAction`, it seems both the typing could be simpler (e.g. no `Omit` is needed anymore).

ocavue:
Your change for `setUploadPlaceholderAction` looks good to me. I'm thinking maybe we can move it forward and add multiple files uploading feature to [`uploadFile`](https://github.com/remirror/remirror/blob/6da3702ba1cb0a138798c8c715f79fe98c93d469/packages/remirror__core/src/builtins/upload-extension/file-upload.ts#L41), and maybe also relative places in the [@remirror/extension-file](https://github.com/remirror/remirror/blob/6da3702ba1cb0a138798c8c715f79fe98c93d469/packages/remirror__extension-file/src/file-extension.tsx#L164-L173). After all, `setUploadPlaceholderAction` is most like an internal API and most people won't use it directly.

packages/remirror__core/src/builtins/upload-extension/file-placeholder-actions.ts
shlroland: Even though the type property in `PlaceholderPluginMeta` doesn't have much use, keeping it directly can make the code look a bit simpler. I'll have a try.

shlroland:
> Your change for `setUploadPlaceholderAction` looks good to me. I'm thinking maybe we can move it forward and add multiple files uploading feature to [`uploadFile`](https://github.com/remirror/remirror/blob/6da3702ba1cb0a138798c8c715f79fe98c93d469/packages/remirror__core/src/builtins/upload-extension/file-upload.ts#L41), and maybe also relative places in the [@remirror/extension-file](https://github.com/remirror/remirror/blob/6da3702ba1cb0a138798c8c715f79fe98c93d469/packages/remirror__extension-file/src/file-extension.tsx#L164-L173). After all, `setUploadPlaceholderAction` is most like an internal API and most people won't use it directly.

I agree with your perspective. It seems like we need to separate the functionality into three methods: `add` `remove` and `update`, to make them directly available for users to use.
The current  [`uploadFiles`](https://github.com/remirror/remirror/blob/6da3702ba1cb0a138798c8c715f79fe98c93d469/packages/remirror__extension-file/src/file-extension.tsx#L164-L173) is still functional, treating each file individually for uploading. If I understand correctly, you're asking if we should consider it as an array of files in the file-upload to achieve the uploading functionality, is that right?

ocavue:
> The current [uploadFiles](https://github.com/remirror/remirror/blob/6da3702ba1cb0a138798c8c715f79fe98c93d469/packages/remirror__extension-file/src/file-extension.tsx#L164-L173) is still functional, treating each file individually for uploading. If I understand correctly, you're asking if we should consider it as an array of files in the file-upload to achieve the uploading functionality, is that right?

Current `uploadFiles` already support multiple files, by dispatching multiple individual transactions. With your change in this PR, we can only achieve that within one single transaction, which could bring better performance more or less. 

> I agree with your perspective. It seems like we need to separate the functionality into three methods: add remove and update, to make them directly available for users to use.

By using `uploadFiles`, the editor will trigger `AddPlaceholderAction` and `RemovePlaceholderAction` (when the uploading is done). Speaking of that, I would like to know in which scene people would want to use `UpdatPlaceholderAction`. We should write done some kind of document, unit test, or a storybook story to show such scene.

shlroland:
> By using uploadFiles, the editor will trigger AddPlaceholderAction and RemovePlaceholderAction (when the uploading is done). Speaking of that, I would like to know in which scene people would want to use UpdatPlaceholderAction. We should write some kind of document, unit test, or a storybook story to show such scene.

I noticed your PR #2100. This pr makes The modifications in this PR caused some of my current test cases to fail. So I think there is a tight coupling between the current upload-extension and file-extension, so I believe that the file-placeholder-plugin should exist as a plugin for file-extension rather than being a built-in extension.In this case, the file-placeholder-plugin can be used as a specific placeholder for the file-extension, while users can utilize the relevant methods from the decoration-extension when they need to use general placeholder functionality.

shlroland:
> Speaking of that, I would like to know in which scene people would want to use UpdatPlaceholderAction.
In my project, the use case for the upload-placeholder is in scenarios where asynchronous requests are needed. A typical example is when I need to call a backend API to re-sign copied resource nodes, ensuring that the permissions of the resource nodes are not compromised.

shlroland:
It's worth noting that currently, the [uploadImage](https://github.com/remirror/remirror/blob/4fdaffd9caa2aee2eb8c0a1308572ce7e7a0b47c/packages/remirror__extension-image/src/image-extension.ts#L177) in the `image-extension` is calling the [placeholder method](https://github.com/remirror/remirror/blob/4fdaffd9caa2aee2eb8c0a1308572ce7e7a0b47c/packages/remirror__core/src/builtins/decorations-extension.ts#L98C67-L98C91) within the `decoration-extension`. When building rich text editor, the upload interfaces for images and files might often be the same, but they require adapting to two different upload logics, which can be confusing. Perhaps we should find a way to unify these two upload logics while keeping compatibility.

PR #2097, opened 2023-06-26
feat(floating-menu): replace popperjs with floatingui

shlroland:
#### Features
- add `middleware` and `strategy ` props to `useFloating` hook.
- add `useFloatingPortal` to portal floatingElement to specified container 

#### Breaking change
- ~~The placement options prefixed with auto- are no longer supported because the placement feature in floatingui no longer supports them. This is primarily due to a conflict between autoPlacement and flip in floatingui. 
https://floating-ui.com/docs/autoPlacement~~


packages/remirror__react-components/src/floating-menu.tsx
```diff
@@ -33,13 +40,6 @@ interface BaseFloatingPositioner extends UseEditorFocusProps {
    */
   hideWhenInvisible?: boolean;
 
-  /**
-   * Set animated as detailed [here](https://reakit.io/docs/popover/#animating).
-   *
-   * Currently this is turned off due to problems with an infinite loop.
-   */
-  animated?: boolean | number;
-
```
ocavue, line 42: I don't think we should remove any option in a non-major version, unless this option is never been used at all.

shlroland: okay.I'll remove the commit.

ocavue:
> The placement options prefixed with auto- are no longer supported

Thank you for making this PR. I would say that replacing PopperJS with FloatingUI is a good idea in general. I have been using FloatingUI in multiple projects and I'm really happy about it. 

However, since the current Remirror API is directly based on PopperJS (i.e. the [placement](https://github.com/remirror/remirror/blob/6da3702ba1cb0a138798c8c715f79fe98c93d469/packages/remirror__react-components/src/floating-menu.tsx#L48-L51) prop), I don't think we can replace PopperJS without breaking backward compatibility. We can add FloatingUI in next major version of Remirror, which doesn't have a clear timeline yet. 

For now, if your like FloatingUI, you can just implement your own `<FloatingPositioner/>` component and don't use the one from Remirror. This could give your more flexibility anyway.

shlroland:
> For now, if your like FloatingUI, you can just implement your own `<FloatingPositioner/>` component and don't use the one from Remirror. This could give your more flexibility anyway.

Yes, I have already implemented it that way.

> However, since the current Remirror API is directly based on PopperJS (i.e. the [placement](https://github.com/remirror/remirror/blob/6da3702ba1cb0a138798c8c715f79fe98c93d469/packages/remirror__react-components/src/floating-menu.tsx#L48-L51) prop), I don't think we can replace PopperJS without breaking backward compatibility. We can add FloatingUI in next major version of Remirror, but I don't have a clear timeline for it yet.

Indeed, it's not a good idea to break existing compatibility at this point. However, I still want to add the functionality of transferring the floating menu to a specific container, so I suggest submitting a new PR based on the current popperjs implementation and mark the current PR as a draft.

shlroland:
> > The placement options prefixed with auto- are no longer supported
> 
> Thank you for making this PR. I would say that replacing PopperJS with FloatingUI is a good idea in general. I have been using FloatingUI in multiple projects and I'm really happy about it.
> 
> However, since the current Remirror API is directly based on PopperJS (i.e. the [placement](https://github.com/remirror/remirror/blob/6da3702ba1cb0a138798c8c715f79fe98c93d469/packages/remirror__react-components/src/floating-menu.tsx#L48-L51) prop), I don't think we can replace PopperJS without breaking backward compatibility. We can add FloatingUI in next major version of Remirror, but I don't have a clear timeline for it yet.
> 
> For now, if your like FloatingUI, you can just implement your own `<FloatingPositioner/>` component and don't use the one from Remirror. This could give your more flexibility anyway.



> > The placement options prefixed with auto- are no longer supported
> 
> Thank you for making this PR. I would say that replacing PopperJS with FloatingUI is a good idea in general. I have been using FloatingUI in multiple projects and I'm really happy about it.
> 
> However, since the current Remirror API is directly based on PopperJS (i.e. the [placement](https://github.com/remirror/remirror/blob/6da3702ba1cb0a138798c8c715f79fe98c93d469/packages/remirror__react-components/src/floating-menu.tsx#L48-L51) prop), I don't think we can replace PopperJS without breaking backward compatibility. We can add FloatingUI in next major version of Remirror, but I don't have a clear timeline for it yet.
> 
> For now, if your like FloatingUI, you can just implement your own `<FloatingPositioner/>` component and don't use the one from Remirror. This could give your more flexibility anyway.

That's a good suggestion as well. We can create a new set of components using floatingui and mark the current floating-menu component as deprecated, planning to remove it in the next major version. This way, we can introduce the new components gradually while providing users with a clear migration path.

shlroland:
> However, since the current Remirror API is directly based on PopperJS (i.e. the [placement](https://github.com/remirror/remirror/blob/6da3702ba1cb0a138798c8c715f79fe98c93d469/packages/remirror__react-components/src/floating-menu.tsx#L48-L51) prop), I don't think we can replace PopperJS without breaking backward compatibility. We can add FloatingUI in next major version of Remirror, but I don't have a clear timeline for it yet.

That's a good suggestion as well. We can create a new set of components using floatingui and mark the current floating-menu component as deprecated, planning to remove it in the next major version. This way, we can introduce the new components gradually while providing users with a clear migration path.

ocavue:
Yep I think adding a new component based on floating-ui is a better idea for compatibility. Maybe we can add `@floating-ui/react` as a peerDependency of the package `@remirror/react-components`, so that we don't increase the package size.

atomiks:
You can keep `placement` backwards-compatible by doing something like this:

```js
placement: placement === 'auto' ? undefined : placement,
middleware: [
  placement === 'auto' ? autoPlacement() : flip()
],
```

If you currently don't expose any internals of Popper like `modifiers`, then it should be possible to upgrade without breaking changes. 

By wrapping the underlying implementation with an abstraction, you can upgrade to new major versions of Floating UI without having to release a new major whenever it has a new major.

shlroland:
> You can keep `placement` backwards-compatible by doing something like this:
> 
> ```js
> placement: placement === 'auto' ? undefined : placement,
> middleware: [
>   placement === 'auto' ? autoPlacement() : flip()
> ],
> ```
> 
> If you currently don't expose any internals of Popper like `modifiers`, then it should be possible to upgrade without breaking changes.
> 
> By wrapping the underlying implementation with an abstraction, you can upgrade to new major versions of Floating UI without having to release a new major whenever it has a new major.

This could be a good approach to keep  backwards-compatible. @ocavue  What do you think?

ocavue:
@atomiks Thanks for your suggestion!

@shlroland Let's try atomiks's approach in this PR

ocavue (approved):
Looks good to me!

PR #2128, opened 2023-07-20
feat(core): decouple Lingui from core, move into optional `@remirror/i18n`

whawker:
### Description
Based on community feedback, we have decided to decouple the core of Remirror from Lingui, an internationalisation (a.k.a. i18n) library.

Thereby making it possible to use _any_ i18n solution with Remirror 🙌🙌🙌.

**N.B.** To use the translatable strings provided by Remirror, the i18n library you use needs to support [ICU message formatting](https://formatjs.io/docs/core-concepts/icu-syntax/).

This change aims to make it **easier to use Remirror in existing applications**, by not imposing _our_ architectural decisions on to you.

There are example integrations with many different i18n libraries [in our Storybook](https://pr2128-remirror-ocavue.vercel.app/?path=/story/i18n-format-js--basic).

### NOTE: "Out-of-the-box" editors unaffected

If you are using editors provided by the `@remirror/react-editors` package, you are unaffected by these changes. These editors have been updated to keep existing behaviour.

## `i18n` prop removed from the `<Remirror />` root component

In previous versions of Remirror, the `i18n` prop of the root Remirror component allowed you to pass a customised **Lingui** instance.

With this version, we want to allow _**any**_ i18n library to be used with Remirror, so the `i18n` prop has been removed, and **replaced with an `i18nFormat` _function_**.

This allows users to plug in _any_ i18n library, by implementing a definition for this function.

This function is described by the TypeScript type [`I18nFormatter`](https://github.com/remirror/remirror/blob/32d8d00587f2f0bce8c1fa59164e15b3569a7e96/packages/remirror__core-types/src/core-types.ts#L417-L453).

#### Example: Using `react-i18next`

```tsx
import { useTranslation } from 'react-i18next';
import { Remirror, useRemirror } from '@remirror/react';

const Editor: React.FC = () => {
  const { t } = useTranslation();

  const i18nFormat: I18nFormatter = useCallback(
    (message, values) => {
      // Note only using the message ID here, more on this later
      return t(message.id, values);
    },
    [t],
  );

  const { manager } = useRemirror({
    extensions: () => [
      // Some extensions here
    ],
  });

  return <Remirror manager={manager} i18nFormat={i18nFormat} />;
};
```

`react-i18next`, like many i18n solutions, requires you define your translatable strings up front, via key-value pairs.

To facilitate this, the `@remirror/messages` package **now exposes the translatable strings as JSON files**.

These messages are provided as key value pairs, so they can be loaded into your chosen i18n library.

Currently, only English locale (`en`) messages are provided.

```ts
import i18n from 'i18next';
import ICU from 'i18next-icu';
import { initReactI18next } from 'react-i18next';
import type { I18nFormatter } from 'remirror';
import allMessages from '@remirror/messages/en/all-messages.json';

// or messages for specific extension(s)
// import boldMessages from '@remirror/messages/en/extension-bold-messages.json';
// import italicMessages from '@remirror/messages/en/extension-italic-messages.json';

i18n
  .use(ICU) // Required if using the provided messages from @remirror/messages
  .use(initReactI18next)
  .init({
    resources: {
      en: {
        translation: allMessages,
      },
    },
    lng: 'en',
    fallbackLng: 'en',
    interpolation: {
      escapeValue: false,
    },
  });
```

You do not _have_ to use the messages in these key value pairs, you could replace them with your own. They are provided for convenience, and to expose the message IDs Remirror uses.

### Restoring the previous behaviour

If you wish to carry on using Remirror's default i18n solution (powered by Lingui), **you will need to install the `@remirror/i18n` package, as this is now an optional package**.

#### Example: Continue using `@remirror/i18n`

Install the i18n package, as it is now optional, and not installed by default.

```sh
npm add @remirror/i18n
```

```tsx
import { i18nFormat } from '@remirror/i18n';
import { Remirror, useRemirror } from '@remirror/react';

const Editor: React.FC = () => {
  const { manager } = useRemirror({
    extensions: () => [
      // Some extensions here
    ],
  });

  return <Remirror manager={manager} i18nFormat={i18nFormat} />;
};
```

## The `useI18n` hook has a different return value

As a consequence of the above, the `useI18n` no longer returns an _object_ containing the Lingui `i18n` instance.

It now returns the `i18nFormat` _function_ that was passed to the root `<Remirror />` component.

#### Before

```tsx
const { t, i18n } = useI18n();
```

#### After

```tsx
const t = useI18n();

// Where "t" is the same function that was passed via `i18nFormat`
```

### Checklist


ocavue (approved):

ocavue:
@whawker Could you rebase this branch so that the CI can run?

packages/remirror__i18n/lingui.config.js
```diff
@@ -1,6 +1,6 @@
 const path = require('path');
 
-const rootDir = path.resolve(__dirname);
+const rootDir = path.resolve(path.join(__dirname, '..', '..'));
 
 module.exports = {
```
ocavue, line 5: `packages/remirror__i18n/lingui.config.js` is a CommonJS file, but the nearest `package.json`, which is `packages/remirror__i18n/package.json` has `"type": "module"`. This will let Node.JS treat `lingui.config.js` as an ESM file. This could (or could not) cause issues. 

I think we have three options here:

1. Make `lingui.config.js` as an ESM file, by using `export default` instead of `module.exports`, if LingUI supports ESM config file.
2. Rename `lingui.config.js` to `lingui.config.cjs`, so that it always be a CommonJS file, if LingUI supports `.cjs` config file extension. 
3. Remove the `"type": "module"` from the `remirror__i18n/package.json` and make it as a CommonJS package.

packages/remirror__messages/src/extension-bold-messages.ts
```diff
@@ -1,14 +1,13 @@
-import type { MessageDescriptor } from '@lingui/core';
-import { defineMessage } from '@lingui/macro';
+import type { RemirrorMessage } from '@remirror/core-types';
 
-export const LABEL: MessageDescriptor = defineMessage({
+export const LABEL: RemirrorMessage = /*i18n*/ {
```
ocavue, line 3: Why do we need `/*i18n*/` here?

packages/remirror__messages/src/index.ts
```diff
@@ -27,4 +27,3 @@ export * as ExtensionTextHighlightMessages from './extension-text-highlight-mess
 export * as ExtensionUnderlineMessages from './extension-underline-messages';
 export * as ExtensionWhitespaceMessages from './extension-whitespace-messages';
 export * as ReactComponentMessages from './react-components-messages';
-export type { MessageDescriptor } from '@lingui/core';
```
ocavue, line 30: We should add a changeset for `@remirror/messages` since its public API has changed. If the only thing we remove is a TypeScript type, I guess we could bump a just minor version for this package.

packages/storybook-react/stories/i18n/remirror-i18n.stories.tsx
```diff
@@ -0,0 +1,5 @@
+import Basic from './remirror-i18n';
+
+export default { title: 'i18n / @remirror⁄i18n' };
```
ocavue, line 3: The character between `@remirror` and `i18n` is weird.

packages/remirror__messages/scripts/messages-to-json.ts
```diff
@@ -0,0 +1,57 @@
+/* eslint-env node */
```
ocavue, line 1: Could you explain a bit why we need all these JSON files? I must miss something.

packages/remirror__messages/src/index.ts
whawker: Yep, haven't added a changeset at all yet. But to be honest I was assuming _everything_ would be major bumped to avoid confusion?

packages/remirror__messages/scripts/messages-to-json.ts
ocavue: I know. These JSON files are part of the public API in package.json

packages/storybook-react/stories/i18n/remirror-i18n.stories.tsx
whawker: Yeah, unfortunately adding a normal slash character makes Storybook split it into a sub folder. It seems Storybook no longer allows you to change this.

packages/remirror__i18n/lingui.config.js
ocavue: I can work on this one if you'd like me to help

packages/remirror__messages/scripts/messages-to-json.ts
whawker: This is so other i18n libraries can consume the messages, without having to load `@remirror/i18n` and hence Lingui

packages/remirror__messages/src/extension-bold-messages.ts
whawker: This is how you can mark a message for Lingui extraction, without using their `defineMessaage` API

https://lingui.dev/guides/message-extraction#explicitly-marking-messages

packages/remirror__i18n/lingui.config.js
whawker: I'll investigate this one 👍

packages/remirror__messages/src/index.ts
whawker: No, I think we need to bump everything because I've changed the API on the root `Remirror` React component, it now expects an `i18nFormat` function instead of a `i18n` prop which is a Lingui instance.

Even if we kept the prop name the same the prop value itself is completely different

packages/remirror__i18n/lingui.config.js
whawker: Moved this functionality back up to the root of the monorepo.

ocavue (approved):
Looks good to me. You can merge it to `v3` if you're happy about it. 

By the way, I merged the latest `origin/main` into `v3` and caused this PR to have some conflict.

PR #2131, opened 2023-07-22
refactor: move @linaria/core dependency from core-helpers to core

p-kuen:
### Description

I am using the tiptap editor which uses `prosemirror-trailing-node` and therefore `@remirror/core-helpers` package, but not `@remirror/core` package.
This PR moves one function which uses the `@linaria/core` package to the core package to reduce dependencies for core-helpers as this package is often used without remirror/core.

I searched all remirror packages where the `cx` function was used and it was always used from the `core` package, so this PR should not have much impact on other packages. Still, as the function was exported, this is possibly a breaking change.

### Checklist

### DIFF
--- a/.changeset/polite-kids-unite.md
+++ b/.changeset/polite-kids-unite.md
@@ -1,0 +1,6 @@
+---
+'@remirror/core-helpers': major
+'@remirror/core': patch
+---
+
+move cx function from core helpers to core package
--- a/.changeset/polite-kids-unite.md
+++ b/.changeset/polite-kids-unite.md
@@ -1,6 +1,7 @@
 ---
 '@remirror/core-helpers': major
 '@remirror/core': patch
+'remirror': patch
 ---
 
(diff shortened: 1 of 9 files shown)


p-kuen:
I removed the exclamation mark from the PR title as this was the reason for a failed check.

ocavue:
Thanks for your contribution. However, I think it's ok to let `@linaria/core` stay in the `@remirror/core-helpers`. One reason is that, as you mentioned, it's a breaking change. Another reason is that the extra dependencies won't have any actual effect if you didn't import it, as the bundlers nowaday should be able to to the tree-shaking for you.

p-kuen:
Thanks for your comment. I understand your point, but I still think that a whole css framework library with the complete babel framework as a dependency for a helper package is not ideal in my opinion and would fit much better in the core package. If you disagree, please close this PR.
Thanks!

PR #2132, opened 2023-07-22
fix: plugins of tableExtension can be updated when editable has been changed

shlroland:
### Description

The table plugin can not be updated when the editable state changes.So I add an editable state change event and use it into tableExtension.
I also believe that more plugins may need to respond correctly to changes in the editable state.

### Checklist

### DIFF
--- a/packages/remirror__extension-events/src/events-extension.ts
+++ b/packages/remirror__extension-events/src/events-extension.ts
@@ -176,6 +176,12 @@ export interface EventsOptions {
    * which was hovered at the current position.
    */
   hover?: Handler<HoverEventHandler>;
+
+  /**
+   * Listen for editable changed and pass through previous editable state and
+   * current editable state
+   */
+  editable?: Handler<EditableEventHandler>;
 }
 
 export type FocusEventHandler = (event: FocusEvent) => boolean | undefined | void;
@@ -204,6 +210,7 @@ export type HoverEventHandler = (
   event: MouseEvent,
   state: HoverEventHandlerState,
 ) => boolean | undefined | void;
+export type EditableEventHandler = (currentEditable: boolean) => void;
 
 /**
  * The events extension which listens to events which occur within the
@@ -233,6 +240,7 @@ export type HoverEventHandler = (
     'copy',
     'cut',
     'paste',
+    'editable',
   ],
   handlerKeyOptions: {
     blur: { earlyReturnValue: true },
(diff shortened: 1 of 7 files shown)


pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
ocavue, line 1: Let's don't commit `pnpm-lock.yaml` if you don't change any dependencies. You can use `corepack enable` to make sure that you have the same PNPM version as described in the [package.json](https://github.com/remirror/remirror/blob/30ff3ddc073125fac7f7a354802dcbea57187075/package.json#L207)

packages/remirror__extension-events/src/events-extension.ts
```diff
@@ -176,6 +176,12 @@ export interface EventsOptions {
    * which was hovered at the current position.
    */
   hover?: Handler<HoverEventHandler>;
+
+  /**
+   * Listen for editable changed and pass through previous editable state and
+   * current editable state
+   */
+  editable?: Handler<EditableEventHanlder>;
```
ocavue, line 184:
```suggestion
  editable?: Handler<EditableEventHandler;
```

packages/remirror__extension-events/src/events-extension.ts
```diff
@@ -204,6 +210,10 @@ export type HoverEventHandler = (
   event: MouseEvent,
   state: HoverEventHandlerState,
 ) => boolean | undefined | void;
+export type EditableEventHanlder = (
+  previousEditable: boolean,
+  editable: boolean,
+) => boolean | undefined | void;
```
ocavue, line 216: I guess we only need to accept one boolean value, since the other one is always the opposite. I prefer the current editable value.

ocavue: Seems like this function doesn't need to return a boolean value. It won't get used anyway.

ocavue:
Thanks for your contribution. I left some comments. Also, could you add the changeset files for the three changed packages? You can learn more from the changeset-bot comment in this PR.

packages/remirror__extension-events/src/events-extension.ts
```diff
@@ -204,6 +210,7 @@ export type HoverEventHandler = (
   event: MouseEvent,
   state: HoverEventHandlerState,
 ) => boolean | undefined | void;
+export type EditableEventHandler = (currentEditable: boolean) => boolean | undefined | void;
```
ocavue, line 213: I guess we don't need to return a boolean value here.

shlroland: > We don't need to return a boolean value here, as the framework doesn't check the return value for this handler.

sorry, I forgot to make changes here

ocavue (approved):

PR #2145, opened 2023-08-02
docs: correct link in extension.md

sorenbs:
This is a simple docs change to fix a broken link on this page: https://remirror.io/docs/concepts/extension/

Clicking the link in Chrome on Macos leads to a 404. Refreshing that 404 then redirects to the same URL without the trailing `/index`. So I assume removing that part will fix the 404.


ocavue:
Thanks!

PR #2169, opened 2023-10-06
feat: remirror V3

ocavue:
### Checklist


ocavue:
@whawker FYI, you can already install `remirror@3.0.0-beta.0`

packages/remirror__core/src/extension/extension-decorator.ts
```diff
@@ -154,7 +154,14 @@ export type ExtensionDecoratorOptions<Options extends Shape = EmptyShape> = Defa
...
   options: ExtensionDecoratorOptions<Options>,
 ) {
-  return <Type extends AnyExtensionConstructor>(ReadonlyConstructor: Type): Type => {
+  return <Type extends AnyExtensionConstructor>(
+    ReadonlyConstructor: Type,
+    context: ClassDecoratorContext<Type>,
+  ): Type => {
+    if (context.kind !== 'class') {
+      throw new Error(`The extension decorator can only be used on a class.`);
+    }
```
whawker, line 163: [In the docs](https://remirror.io/docs/concepts/extension#extension-1) we mention that you can use decorators as a normal function

Screenshot 20
23-11-02 at 18 13 35

For this to be true now, we would need to use (at a minimum)

```diff
- extension({ defaultSettings: { color: 'red' } })(ExampleExtension);
+ extension({ defaultSettings: { color: 'red' } })(ExampleExtension, { kind: 'class' });
```

Should we let `context` default to `{ kind: 'class' }` if undefined? Or should we update the docs instead?

ocavue: we let context default to `{ kind: 'class' }`

whawker: Should an appropriate default be added for all decorators, or just this `extension` one?

ocavue: I think all decorators from Remirror should have default `kind`

PR #2191, opened 2023-11-03
feat: make `extension` decorator backwards compatible when called as function

whawker:
### Description

💥 BREAKING 💥: Remove deprecated decorator `extensionDecorator`. Please use `extension` instead

Make `extension` decorator backwards compatible when called directly as a function.

### Checklist

### DIFF
--- a/.changeset/smart-garlics-bow.md
+++ b/.changeset/smart-garlics-bow.md
@@ -1,0 +1,8 @@
+---
+'@remirror/core': patch
+'remirror': patch
+---
+
+💥 BREAKING 💥: Remove deprecated decorator `extensionDecorator`. Please use `extension` instead
+
+Make `extension` decorator backwards compatible when called directly as a function.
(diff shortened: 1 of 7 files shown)


ocavue (approved):
LGTM. We could also support undefined context for other decorators like [`@command`](https://github.com/remirror/remirror/blob/8f5467ae62d81c900e371edee13da7bf7c6895ad/packages/remirror__core/src/builtins/builtin-decorators.ts#L146-L150) etc 


https://github.com/remirror/remirror/blob/8f5467ae62d81c900e371edee13da7bf7c6895ad/packages/remirror__core/src/builtins/builtin-decorators.ts#L146-L150

whawker:
`@command`, `@helper` and `@keyBinding` use `context.addInitializer`, so we can't make `context` optional here.

Calling these decorators directly as a "function" was never included as part of the public API in my opinion, whereas using `extension` _was_ mentioned in the docs.

PR #2237, opened 2024-02-19
fix(markdown): apply html sanitizer on markdowntohtml output

Pigment-RomainLoisel:
### Description

There is a bug in code, `htmlSanitizer` is never used.
An option should be sent to `marked` with `sanitize: true` to enable it.
Also `marked` [deprecated](https://github.com/markedjs/marked/blob/1c2593f81e7cae4349a84d314ab0eed0ccb99f09/docs/USING_ADVANCED.md?plain=1#L69) and removed the html sanitizer options, it is expected to use sanitizer on `marked` output now in the latest version.
I propose a fix to sanitize the output instead to be futureproof.

### Checklist

    - **There is actually a test that fails, but it also fails on master**


ocavue:
Thanks! I removed the default sanitizer and fixed the tests.

PR #2255, opened 2024-05-26
feat: upgrade to prettier v3.2

### LINKED ISSUE #2250: Remove prettier dependency
### Summary

When i try to install packages with pnpm and have prettier in my package.json as a dependency i get a warning from a package manager that the versions are unmet.

### Steps to reproduce

```pnpm add prettier && pnpm add remirror @remirror/react```

### Expected results

no warning when trying to add with pnpm

### Actual results

 WARN  Issues with peer dependencies found
.
├─┬ @remirror/react 2.0.35
│ └─┬ @remirror/extension-react-tables 2.2.19
│   └─┬ @remirror/react-components 2.1.17
│     └─┬ @remirror/extension-code-block 2.0.18
│       └── ✕ unmet peer prettier@^2.8.8: found 3.2.5
└─┬ remirror 2.0.39
  └── ✕ unmet peer prettier@^2.8.8: found 3.2.5

### PR DESCRIPTION
**Pull request by mdmower**
### Description

- Update prettier throughout project to v3.2.5 and set peer dependency to `^3.2.0`.
- Remove \@types/prettier dependencies since types are bundled in prettier v3.
- Update `fix:prettier` npm script to use `--log-level` argument.
- Add prettier as a dependency in `packages/remirror__cli` since it is used directly in source.
- Update `packages/remirror__extension-code-block` to run formatter asynchronously.
  - Update corresponding unit tests to let formatter apply asynchronously before comparing before/after content.
  - Add additional test to make sure content is unchanged on format failure.
- Update `formatter` extension in `packages/remirror__extension-code-block` to use prettier v3.
- Update `support/scripts/src/linaria.ts` to fix prettier config resolution which needs to be based on file path, not directory.
- Format source with updated prettier.

#### Breaking change

This PR is categorized as **feat:** because it introduces a breaking change on the `formatter` prop that is accepted by the code block extension. The signature changed from
```
formatter: (params: FormatterProps) => FormattedContent | undefined
```
to
```
formatter: (params: FormatterProps) => Promise<FormattedContent | void>
```

### Checklist

### Screenshots

No user facing changes. A new code formatter entry in the storybook is introduced in separate PR #2256 that depends on this PR. Screenshots of the formatter in action are available there.

Fixes #2250

### CODE REVIEW

**Comment by mdmower** [comment]
@ocavue - I know the prettier v3 update was [attempted before](https://github.com/remirror/remirror/pull/2139) and then [reverted](https://github.com/remirror/remirror/pull/2149), but I've taken a different approach here (async formatting). Would you be willing to review? In PR #2256 (that depends on this one), I've introduced a storybook entry where you can test the updated formatter.

**Comment by mdmower** [comment]
@whawker @ocavue - I've rebased this PR and the code formatter ~~extension~~ component PR #2256 on branch `v3`. It looks like the `pr-name-linter` CI check is failing, but that seems like something where you two could choose the right names for these PRs. Let me know how else I can help here.

File: packages/remirror__extension-code-block/src/code-block-extension.ts [hunk 1]
```diff
@@ -293,6 +293,10 @@ export class CodeBlockExtension extends NodeExtension<CodeBlockOptions> {
    * Format the code block with the code formatting function passed as an
    * option.
    *
+   * The formatter runs asynchronously. This should not be immediately followed
+   * by additional commands. It is not yet possible to track completion of this
+   * command; it is best to use as a standalone response to a user action.
+   *
```
whawker, line 299: If this is an async command, we should use the `DelayedCommand` interface instead, as this supports promises.

There's an example of [its usage here](https://github.com/remirror/remirror/blob/a0cc38df84d3de7d3fd72ae1804803bf4b76b0de/packages/remirror__core/src/builtins/decorations-extension.ts#L599)

This should mean the command is chainable, but if that's not the case we could alternatively

```ts
@command({
  disableChaining: true,
})
```

packages/remirror__extension-code-block/package.json
```diff
@@ -59,19 +59,14 @@
...
   "peerDependencies": {
     "@remirror/pm": "^3.0.0-beta.5",
-    "@types/prettier": "^2.7.2",
-    "prettier": "^2.8.8"
+    "prettier": "^3.2.0"
   },
   "peerDependenciesMeta": {
-    "@types/prettier": {
-      "optional": true
-    },
```
whawker, line 74: I think this peer dep should still be optional for 2 reasons

As we have a `formatter` option here, the user might be using something _other_ than prettier.  

I suspect this might also result needing to have prettier installed, whether you use the code block extension or not.

packages/remirror__extension-code-block/src/formatter.ts
```diff
@@ -110,37 +112,52 @@ export function formatter(props: FormatterProps): FormattedContent | undefined {
     return { ...result, cursorOffset: result.cursorOffset + increment };
   };
 
+  let parser: BuiltInParserName;
+
+  switch (language) {
+    case 'typescript':
+    case 'ts':
+    case 'tsx':
+      parser = 'typescript';
```
whawker, line 121: Could we pull this out into a separate helper function, then use `return` instead of `break`?

i.e.

```ts
switch (language) {
    case 'typescript':
    case 'ts':
    case 'tsx':
      return 'typescript';
```

whawker (commented):
This PR make sense from a concept point of view, however I think we'd need to refactor to use the `DelayedCommand` interface to fit in with the rest of the ecosystem.

Happy to assist with this of course.

packages/remirror__extension-code-block/src/code-block-extension.ts
mdmower: Thanks for the example. I followed it pretty closely for the changes in [5a498ef](https://github.com/remirror/remirror/pull/2255/commits/5a498ef8009d19c7e0a22d678f5ec97e1ad8a0f8). I'm not familiar with best/worst practices in this project, so I'm curious about whether it's ok to pass an `EditorState` to the promise as I've done in this commit.

While working on these changes, I think I may have fixed a bug [here](https://github.com/remirror/remirror/blob/08e0da14d48f144e8be1c43fea398cfad8f518e7/packages/remirror__extension-code-block/src/code-block-utils.ts#L276-L279). A `pos` parameter is allowed to be passed into the format function to identify a different code block than currently selected, but then the current selection was used to find the code block anyways (i.e. `pos` was ignored). Let me know if you agree with the change I made [here](https://github.com/remirror/remirror/pull/2255/commits/5a498ef8009d19c7e0a22d678f5ec97e1ad8a0f8#diff-72c26535e24a6bdefe72fe7bc93b77f5bf07518fb41d15d8586ce67b7abc4909R293-R296) to utilize `pos` if provided.

Finally, my changes in `packages/remirror__extension-code-block/__tests__/code-block-extension.spec.ts` remain because I'm not sure how else to wait for the command to signal completion before checking the updated doc. Let me know if there's a mechanism for this.

packages/remirror__extension-code-block/package.json
mdmower: I'm not sure I follow. I've only removed `@types/prettier` here because types are included in prettier v3. `prettier` remains listed as an optional peer dependency. Do I misunderstand something?

whawker: Ugh, my mistake. I forgot about the type being included with v3 thing.

I need more sleep clearly! 😴

packages/remirror__extension-code-block/src/code-block-extension.ts
```diff
@@ -293,29 +299,47 @@ export class CodeBlockExtension extends NodeExtension<CodeBlockOptions> {
...
+    return this.store
+      .createFormatCodeBlockCommand({
+        promise: () =>
+          formatCode({
+            type: this.type,
+            formatter: this.options.formatter,
+            defaultLanguage: this.options.defaultLanguage,
+            state: this.store.getState(),
+            ...props,
+          }),
```
whawker, line 325: We should clean this bit up, using `props` here is confusing

`props` usually refers to `CommandFunctionProps`, but this is actually command _arguments_

```ts
formatCodeBlock({ pos }: Partial<PosProps> = {}): CommandFunction {
```

We can get the actual _command_ props, from the promise

```ts
      promise: (props) =>
          formatCode({
            type: this.type,
            formatter: this.options.formatter,
            defaultLanguage: this.options.defaultLanguage,
            pos, // The command argument,
            ...props, // CommandFunctionProps
          }),
```

packages/remirror__extension-code-block/src/code-block-extension.ts
```diff
@@ -458,12 +482,39 @@ export class CodeBlockExtension extends NodeExtension<CodeBlockOptions> {
...
   namespace Remirror {
+    interface ExtensionStore {
+      /**
+       * Create delayed command which formats code block contents.
+       */
+      createFormatCodeBlockCommand<Value = any>(
+        props: DelayedFormatCodeBlockProps<Value>,
+      ): DelayedCommand<Value>;
+    }
+
```
whawker, line 517: As mentioned above, I don't think we should add this to the shared store.

packages/remirror__extension-code-block/src/code-block-utils.ts
```diff
@@ -260,82 +261,108 @@ export function codeBlockToDOM(node: ProsemirrorNode, extra: ApplySchemaAttribut
   return ['pre', attributes, ['code', { [LANGUAGE_ATTRIBUTE]: language, style }, 0]];
 }
 
-interface FormatCodeBlockFactoryProps
-  extends NodeTypeProps,
-    Required<Pick<CodeBlockOptions, 'formatter' | 'defaultLanguage'>> {}
+type FormatCodeProps = NodeTypeProps &
+  Required<Pick<CodeBlockOptions, 'formatter' | 'defaultLanguage'>> &
+  Partial<PosProps> & { state: EditorState };
```
whawker, line 266: Following on the the arguments vs command function props mentioned above,

Instead of `& { state: EditorState }` we should use `& CommandFunctionProps`

packages/remirror__extension-code-block/src/code-block-utils.ts
```diff
@@ -260,82 +261,108 @@ export function codeBlockToDOM(node: ProsemirrorNode, extra: ApplySchemaAttribut
...
+ * Format the contents of a selected codeBlock or one located at the provided
+ * position.
  */
-export function formatCodeBlockFactory(props: FormatCodeBlockFactoryProps) {
-  return ({ pos }: Partial<PosProps> = object()): CommandFunction =>
-    ({ tr, dispatch }) => {
-      const { type, formatter, defaultLanguage: fallback } = props;
+export async function formatCode(props: FormatCodeProps): Promise<FormatCodeResult> {
+  const { type, formatter, defaultLanguage: fallback, pos, state } = props;
+  const { selection, doc } = state;
```
whawker, line 291: Again, following on the the arguments vs command function props mentioned above

You will have `tr` available on props now, and we should that instead of `state`

```ts
  const { type, formatter, defaultLanguage: fallback, pos, tr } = props;
  const { selection, doc } = tr;
```

whawker (commented):
Nice work here!

I've made some suggestions which are mainly a clarification of arguments vs command function props.

mdmower:
@whawker - Thanks for reviewing. I believe all feedback has been addressed in [20479ec](https://github.com/remirror/remirror/pull/2255/commits/20479ec7e15788999624825fea6615a2eba6d28a). Happy to revise more if you see anything else.

Could you help with the included changeset? I'm pretty sure it's inaccurate. I know `@remirror/extension-react-language-select` can be removed (accidentally leftover from PR targeting v2), but I'm unsure about whether `@remirror/extension-code-block` `major` is appropriate for `@remirror/extension-code-block` since there is presumably already a v3 major bump incoming for the extension. I'm also unsure about whether source that only received formatting changes should get `patch` bumps.

packages/remirror__extension-code-block/src/formatter.ts
```diff
@@ -6,25 +6,27 @@
...
 import type { BuiltInParserName, CursorOptions, CursorResult } from 'prettier';
-import babelPlugin from 'prettier/parser-babel';
-import graphqlPlugin from 'prettier/parser-graphql';
-import htmlPlugin from 'prettier/parser-html';
-import markdownPlugin from 'prettier/parser-markdown';
-import cssPlugin from 'prettier/parser-postcss';
-import typescriptPlugin from 'prettier/parser-typescript';
-import yamlPlugin from 'prettier/parser-yaml';
-import { default as Prettier } from 'prettier/standalone';
+import babelPlugin from 'prettier/plugins/babel';
```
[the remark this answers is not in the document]
whawker: Good spot!

whawker:
Thanks for highlighting the issue on the changeset, if we update the code block extension itself (and the `remirror` entry point) it should suffice.

I have opted to use the `formatter` in the unit tests, rather than a mock, as it can't hurt to have a little coverage on that functionality.

It also seems this change also means `formatCodeBlock` cannot be chained, as we consume the text, format it, and write it back at future point in time, potentially overwriting any other changes made, so I have made the command non chainable.

mdmower:
> Thanks for highlighting the issue on the changeset, if we update the code block extension itself (and the `remirror` entry point) it should suffice.

Thanks for cleaning this up.

> I have opted to use the `formatter` in the unit tests, rather than a mock, as it can't hurt to have a little coverage on that functionality.

For what it's worth, it wasn't a mock formatter. Rather, there was a spy wrapping a real formatter to ensure it ran the correct number of times and it enabled a means of more tightly tracking the completion of the format command than arbitrary waits. It was more complicated, though, I'll admit.

> It also seems this change also means `formatCodeBlock` cannot be chained, as we consume the text, format it, and write it back at future point in time, potentially overwriting any other changes made, so I have made the command non chainable.

Thank you for following-up on this; it was a miss on my part since you suggested the possibility earlier.

PR #2258, opened 2024-06-02
refactor(jest-prosemirror): append element to DOM before EditorView initialization in `createEditor`

fyvfyv:
### Description

Some `ProseMirror` plugins rely on the editor's outer DOM structure. For instance, if a plugin's [`view`](https://prosemirror.net/docs/ref/#state.PluginSpec.view) method uses something like `view.dom.parentElement`, it can lead to errors during tests because `parentElement` is `null` when the editor is created with `createEditor`. Appending the element to the DOM ensures plugins that depend on the DOM structure can be tested without issues, improving the robustness and reliability of the createEditor utility function.

### Checklist

### DIFF
--- a/.changeset/honest-socks-sip.md
+++ b/.changeset/honest-socks-sip.md
@@ -1,0 +1,5 @@
+---
+'jest-prosemirror': minor
+---
+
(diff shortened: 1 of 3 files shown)


whawker:
Thanks for the contribution @fyvfyv! All our tests pass and your reasoning for the change itself makes sense.

LGTM

PR #2269, opened 2024-07-18
fix(extension-file): allow `fileSize` to be string

ocavue:
### Description

Allow `fileSize` attribute to be string type so that it's compatible with existing data.

### Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,8 +6,8 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true,
-    "source.fixAll.markdownlint": true
+    "source.fixAll.eslint": "explicit",
+    "source.fixAll.markdownlint": "explicit"
   },
 
   // Files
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -6,8 +6,8 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true,
-    "source.fixAll.markdownlint": true
+    "source.fixAll.eslint": "explicit",
+    "source.fixAll.markdownlint": "explicit"
```
ocavue, line 10: Something VSCode changed automatically.

packages/remirror__extension-file/src/file-extension.tsx
```diff
@@ -88,7 +88,7 @@ export class FileExtension extends NodeExtension<FileOptions> {
         url: { default: '', validate: 'string' },
         fileName: { default: '', validate: 'string' },
         fileType: { default: '', validate: 'string' },
-        fileSize: { default: 0, validate: 'number' },
+        fileSize: { default: 0, validate: 'string|number' },
```
ocavue, line 91: Notice that in `parseDOM`, we simply use `const fileSize = anchor.getAttribute('data-filesize');` without converting the value into a number, that means existing `file` nodes having `fileSize` as `string`.

whawker (approved):

